PR #2, opened 2015-12-14
Coco 135

dstack:
@aarmour


aarmour:
@dstack Clicking outside of the component is not working for me when I run this locally. Keyboard navigation also does not appear to be working.

For handling outside clicks, see https://github.com/Pearson-Higher-Ed/compounds/tree/onClickOutside. I borrowed this idea from [react-onclickoutside](https://github.com/Pomax/react-onclickoutside)—implemented without using a mixin.

For keyboard navigation, there is probably a more idiomatic (React) way of implementing, but I need to study the code further to come up with a suggestion.

Also, I don't see any `aria-` attributes.

(cc @aaronkaka @StommePoes)

src/Dropdown.js
```diff
@@ -0,0 +1,85 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import indexOfElement from './util/indexOfElement';
+import dispatchEvent from './util/dispatchEvent';
+import delegateEvents from './util/delegateEvents';
+
+class Dropdown extends React.Component {
```
aarmour: Why are we renaming from Elements?

src/Dropdown.js
```diff
@@ -0,0 +1,85 @@
...
+        <div data-toggle="dropdown-menu"></div>
+        <div className="pe-dropdown-menu__menu-items">
+          {this.props.children}
+        </div>
+      </div>
+  }
+}
+
+export default Dropdown;
+export function DropdownDivider(props) {
```
aarmour: To stay consistent with other multi-part components (i.e., Card), move all of the dropdown menu pieces to `dropdown-menu` directory and keep one component per file: DropdownMenu.js, DropdownMenuDivider.js, DropdownMenuItem.js

StommePoes: <a href="#" data-toggle="dropdown" data-target="demo-dropdown">Toggle Default Dropdown</a>

For toggles I always suggest buttons. Esp an anchor with href="#", clicking on it would normally move focus back to the top of the page (fragment #). Buttons avoid the need for hrefs to be focusable.

docs/src/app/components/containers/components/Dropdowns.js
```diff
@@ -0,0 +1,80 @@
...
+      }
+    ]
+  }
+];
+
+function DropdownsContainer(props) {
+  return (
+    <div>
+      <Demo code={code}>
+      <a href="#" data-toggle="dropdown" data-target="demo-dropdown">Toggle Default Dropdown</a>
```
aarmour: The trigger should be encapsulated in the component. `data-toggle` and `data-target` shouldn't be needed—use a [ref](https://facebook.github.io/react/docs/more-about-refs.html) instead to grab the menu container.

dstack:
@aarmour @StommePoes all comments addressed

wip deployed at http://pearson-compounds-coco-135.surge.sh/#/components/dropdowns

StommePoes:
Any reason not to make the list of links a list of links? Including the separator to be a li with the separator role.

Although this version doesn't have headings and I've seen the other one does, and a heading can't just be thrown into a list, but if that's going to be something devs will want to do, I'd make the dropdowny part a div who wraps (any number of) ul's and headings can then properly "head" those ul's.

I also wonder if a user hitting escape while that button is toggled on needs to be listened for and focus moved back to the button if a menu item is focussed. What happens if I'm in the menu and hit ESC, while I expect to end up back on the button (or where I was if I've tabbed out of it and am further on the page) is either focus to stay where I am (I ended up getting focussed on something hidden after/behind the dropdown) or if I was in the menu and my focusable vanished, back to the toggle button.

In any case what happens now is I end up at the top of the page. That'll suck on longer pages where the button isn't already in the header of the page (which is what this is meant for, now, but as a all-purpose menu thingie it could and should be able to be used anywhere).

Thoughts?

PR #3, opened 2015-12-17
chore(test): configure unit testing

aarmour:
This PR configures automated testing for component structure. It answers the question: "Given a set of props, does the virtual DOM match the expected structure?". This is the approach [recommended](https://discuss.reactjs.org/t/whats-the-prefered-way-to-test-react-js-components/26) by the react team, and does not require a real DOM. I used [this article](https://blog.algolia.com/how-we-unit-test-react-components-using-expect-jsx/) as a guide. Note that there are no tests yet for *behavior* because the existing components don't have any. (Dropdown menu will, though). We can circle back to whether or not we need to be using Karma at a later time.

@dstack @aamirali1 @StommePoes (cc @aamirali1)


src/__tests__/Label-test.js
```diff
@@ -0,0 +1,59 @@
...
+import expect from 'expect';
+import { createRenderer } from 'react-addons-test-utils';
+import expectJSX from 'expect-jsx';
+expect.extend(expectJSX);
+
+import { Label } from '../';
+
+describe('Label', () => {
+
+  beforeEach(function () {
```
aarmour: Note: when using `this`, you **cannot** use ES2015 arrow functions because they lexically bind the `this` value. E.g.,

```
beforeEach(() => {
  // 'this' is undefined
});
```

dstack:
This loos good as is.

PR #6, opened 2016-02-19
fix: Handle IE override of button input type.

aaronkaka:
It appears that IE overrides the input type on button elements.  Is this a valid way to handle this in your opinion, or will we want to style 'submit' buttons differently than any other button?

### DIFF
--- a/src/Button.js
+++ b/src/Button.js
@@ -2,7 +2,8 @@ import React, { PropTypes } from 'react';
 
 const BUTTON_TYPES = {
   PRIMARY: 'primary',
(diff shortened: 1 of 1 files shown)


StommePoes:
I can't test currently but it's probable that restricting ourselves to HTML type attributes and then giving buttons the "button" type would fix this. A button inside a form has a default type of submit, and so I personally tend to set all pushbuttons (buttons who just run scripts) to type=button to prevent any accidental form submission.

But I would like to see for myself what IE is doing exactly first.

PR #7, opened 2016-03-18
fix(deps): Add elements-sdk dependency, AND do not parse autoit for h…

deepak2510:
…ighlight.js: https://github.com/isagalaev/highlight.js/issues/895


deepak2510:
@aaronkaka

PR #8, opened 2016-03-29
Add enzyme; Add tests for Button, Card and Label components

deepak2510:


aaronkaka:
Let's take this opportunity to do a little package.json cleanup:

1. To get in line with npm3, move "react" from "peerDependencies" to "dependencies" (eliminate peerDependencies).
2. Update the "react" dependency to version "~0.14.7".
3. Move the "react-dom" dependency to "dependencies" from "devDependencies" and set to "~0.14.7".
4. Remove redundant "react" from "devDependencies".
5. Determine if "highlight.js" is actually a dependency for the functionality of the code vs a dev dependency. If so, move it to "dependencies" section.

After doing all this, 'rm -rf node_modules' and 'npm install' to test out locally first.

aaronkaka:
Add a check for node 4.2 in the travis.yml, and see if that CIs green.

PR #14, opened 2016-04-18
Upgrade React to v15 and make React and ReactDom external dependencies

deepak2510:


webpack.config.js
```diff
@@ -14,7 +14,8 @@ module.exports = {
         commonjs2: 'react',
         commonjs: 'react',
         amd: 'react'
-      }
+      },
+      "react-dom" : 'ReactDOM'
```
aaronkaka: Go ahead and do the full treatment for ReactDOM as well:
{
      'react-dom': {
        root: 'ReactDOM',
        commonjs2: 'react-dom',
        commonjs: 'react-dom',
        amd: 'react-dom'
      }
    }

PR #16, opened 2016-07-06
feat: calendar component with basic functions and change the start day of calen…

gayankumara:
…dar according to the locale change,

added unit test for calendar functions ,including the test for LocaleUtils , which moment js handle the "start day" of the calendar accordingly locale.

* added additional route (<Route path="calendar/:locale" component={Calendar} /> ,) to be tested by passing locale through the  url, (locale: fr, hi, en-nz ... ) 
http://localhost:8000/#/components/calendar/fr

### DIFF
--- a/docs/src/app/components/containers/Components.js
+++ b/docs/src/app/components/containers/Components.js
@@ -11,7 +11,8 @@ const navData = [
       { title: 'Buttons' },
       { title: 'Card' },
       { title: 'Labels' },
-      { title: 'Dropdowns' }
+      { title: 'Dropdowns' },
+      { title: 'Calendar' }
     ]
   }
 ];
--- a/docs/src/app/components/containers/components/DateUtils.js
+++ b/docs/src/app/components/containers/components/DateUtils.js
@@ -1,0 +1,149 @@
+/**
+ * Return `d` as a new date with `n` months added.
+ * @param {[type]} d
+ * @param {[type]} n
+ */
+export function addMonths(d, n) {
+  const newDate = clone(d);
+  newDate.setMonth(d.getMonth() + n);
+  return newDate;
+}
+
+/**
+ * Clone a date object.
+ *
+ * @param  {Date} d The date to clone
+ * @return {Date} The cloned date
+ */
+export function clone(d) {
+  return new Date(d.getTime());
+}
+
+/**
+ * Return `true` if two dates are the same day, ignoring the time.
+ *
+ * @param  {Date}  d1
+ * @param  {Date}  d2
+ * @return {Boolean}
+ */
+export function isSameDay(d1, d2) {
+  if (!d1 || !d2) {
+    return false;
+  }
+  return d1.getDate() === d2.getDate() &&
+    d1.getMonth() === d2.getMonth() &&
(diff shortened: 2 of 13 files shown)


gayankumara:
re-open ,because of Travis failure .....

StommePoes:
@umahaea @veerabvm can we find out what's failing in the installs? Gayan also saw a fail looking for "chai". @gayanKumera did your unit tests use mocha and chai?

@wyseguyonline pinging you for also looking at this PR.

gayankumara:
@StommePoes , yes unit tests use mocha and chai
seems mocha dependency installed but "chai" does not.

StommePoes:
@gayanKumara don't worry, we'll see what PDA decides for unit tests like these and if mocha+chai is approved it'll be added to dependencies.
Also, I think I found why lax branch needed the "stage-0" dependency: some of the calendar code is using this kind of code:
```
class someThing extends React.Component {
  static propTypes = {
    property: React.PropTypes.string,
    anotherProperty: React.PropTypes.number
  };
  static defaultProps = {
    property: ' ',
    anotherProperty:  2
  };
...
}
```
This is ES7 "property intializers" which means this is not in the set spec (meaning the Javascript committees could change this and make it not work in the future). The ES6 way, where you wouldn't need the "stage-0" dependency, is to move things like PropTypes and defaultProps to *after* the class declaration like this:
```
class someThing extends React.Component {
... stuff
}
someThing.propTypes = {
  property: React.PropTypes.string,
  anotherProperty: React.PropTypes.number
};
someThing.defaultProps = {
  property: ' ',
  anotherProperty:  2
};
```

Since @aaronKaka has said before that he does not want the experimental stuff in the Compounds  repo (for now), I think adjusting these React classes to match ES6 (by moving those "static" lines to after the class declaration) will let Lax be able to remove that dependency from your branch and the Calendar will be able to run in the master branch.

gayankumara:
@StommePoes ,Thanks for pointing it, Here changed the code in ES6 way, as you commented above. 
it works fine , so that I re-based the code and pushed again.

StommePoes:
@gayanKumara I asked @aaronKaka about chai. He said the Compounds repo has some unit testing already set up by another teammember Deepak, and he used mocha and <a href="https://semaphoreci.com/community/tutorials/testing-react-components-with-enzyme-and-mocha
enzyme</a>.

I'm not familiar with either, but do you need help switching your chai files to enzyme? If so it would be good if someone on the team could be available to answer questions etc.

docs/src/app/components/containers/Components.js
```diff
@@ -11,7 +11,8 @@ const navData = [
       { title: 'Buttons' },
       { title: 'Card' },
       { title: 'Labels' },
-      { title: 'Dropdowns' }
+      { title: 'Dropdowns' },
+      { title: 'Calendar' },
```
wyseguyonline: If this is the last item in the array, there shouldn't be a comma at the end.

StommePoes:
@gayankumara (don't look at this until it's tomorrow! sleep!)
can the "babel-preset-stage-0": "^6.3.13", be removed then from package.json?

PR #17, opened 2016-07-07
Calendar accessibility additions

StommePoes:
@wyseguyonline @gayankumara @umahaea ... take a look and see if anything jumps out at you.

- Decided to do last part of DES-348 here since I was playing with compounds (update guidelines)

- I moved the buttons to after the h3 to match the Elements calendar. This meant removing the caption as a separate thing. 
- Changed some roles, added some roles and aria attributes (the h3 may not need the heading role but it should now announce itself when a user clicking a button updates the month).
- Had to remove the {...attributes} from the main render because it was giving me errors, not sure if that is the right thing to do (was getting this error https://facebook.github.io/react/warnings/unknown-prop.html)

TODO: localisation for the title and aria-label attribute strings!
Later TODO: keyboarding for moving through the dates with arrows.

### DIFF
--- a/docs/src/app/components/containers/components/DayPicker.js
+++ b/docs/src/app/components/containers/components/DayPicker.js
@@ -2,6 +2,7 @@ import React, { PropTypes } from 'react';
 import * as Helpers from './Helpers';
 import * as DateUtils from './DateUtils';
 import * as LocaleUtils from './LocaleUtils';
+import uuid from 'uuid';
 
 const keys = {
   LEFT: 37,
@@ -9,20 +10,6 @@ const keys = {
   ENTER: 13,
   SPACE: 32
 };
-
-class Caption extends React.Component { // eslint-disable-line
-
-  render() {
-    const { date, locale, localeUtils, onClick } = this.props;
-    return (
-      <div className="pe-calendar-year">
-      <h3 className="pe-calendar-month" onClick = {onClick} >
-        {localeUtils.formatMonthTitle(date, locale)}
-      </h3>
-      </div>
-    );
-  }
-}
 
 export default class DayPicker extends React.Component {
 
@@ -219,11 +206,6 @@ export default class DayPicker extends React.Component {
     }
   }
 
-  handleCaptionClick(e, currentMonth) {
-    e.persist();
-    this.props.onCaptionClick(e, currentMonth);
-  }
-
   handleDayClick(e, day, modifiers) {
     e.persist();
     if (modifiers.indexOf("outside") > -1) {
@@ -261,7 +243,7 @@ export default class DayPicker extends React.Component {
     const rightButton = isRTL ? this.allowPreviousMonth() : this.allowNextMonth();
 
     return (
-      <div className="pe-calendar-year pe-calendar-buttons">
+      <div className="pe-calendar-buttons">
         {leftButton &&
           <button
             key="left"
@@ -277,9 +259,9 @@ export default class DayPicker extends React.Component {
         {rightButton &&
           <button
             key="right"
-            className={`pe-btn--inverse-link`}
             aria-label="Next"
             title="Next"
+            className={`pe-btn--inverse-link`}
             onClick={() => isRTL ? this.showPreviousMonth() : this.showNextMonth()}
             >
             <span className="pe-icon--chevron-right" aria-hidden="true"></span>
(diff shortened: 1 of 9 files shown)


umahaea:
looks ok to me..

gayankumara:
@StommePoes , seems ok , BTW we need to go through the QE testing and if anything will inform you.
so that we have user story (http://jira.pearsoncmg.com/jira/browse/GCAPI-1335) to test the branch 

Seems there is no any function called for isDayExists(...), in docs/src/app/components/containers/components/DateUtils.js
we can remove it. so what do you think about it?

StommePoes:
@gayanKumara hey also, didn't you have a newer, better way for this:
weekDayShort = moment().locale(locale).weekday(day).format('dd').charAt(0);
 ?
if so let's add it in.

I removed isDayExists

docs/src/app/components/containers/components/DayPicker.js
```diff
@@ -407,7 +397,7 @@ export default class DayPicker extends React.Component {
 
     const { localeUtils, locale } = this.props;
     const ariaLabel = localeUtils.formatDay(day, locale);
-    const ariaDisabled = isOutside ? "true" : "false";
+    const ariaDisabled = isOutside ? 'true' : 'false';
```
wyseguyonline: Is there a reason why these are strings rather than pure booleans?

docs/src/app/components/containers/components/LocaleUtils.js
```diff
@@ -2,36 +2,36 @@ import moment from 'moment';
...
 
 export function formatWeekdayShort(day, locale) {
 
   let weekDayShort;
   if (locale === 'en') {
     // if the loacle is default - english calendar then update as 'spec'
-    weekDayShort = moment().locale(locale).weekday(day).format("dd").charAt(0);
+    weekDayShort = moment().locale(locale).weekday(day).format('dd').charAt(0);
 
   } else {
```
wyseguyonline: Can this value be set when weekDayShort is initialized and then changed if locale equals 'en'?

docs/src/app/components/containers/components/DayPicker.js
StommePoes: I didn't write that part but in HTML they're not booleans (some attributes are real booleans like required or disabled) but  actually have string values (this might have been done because of issues that appeared when aria-hidden started being used in the wild).
However the reason they are strings here is more likely to do with how the original author likes to do things. I only edited this to make J Random Linter happy,

docs/src/app/components/containers/components/LocaleUtils.js
StommePoes: Gayan told me he uses the charAt(0) because the elements design has 1 letter but he showed me on his branch that if you add "/en" (or "/nl") to the end of the URL you get the "dd" version (the "short but not 1 letter" version).

If we switch to Compounds being React Intl then this part all gets changed anyway.

wyseguyonline: Sorry, my point is can we rewrite it like this:

let weekDayShort = moment().locale(locale).weekday(day).format('dd');
if (locale === 'en') {
  weekDayShort = moment().locale(locale).weekday(day).format('dd').charAt(0);
}

StommePoes: Probably. This whole thing could be rewritten and I suspect it could be half the current size for what we need a calendar to do. If you want I can change this, I also personally prefer ifs-branch-off to elses myself.

wyseguyonline:
Code looks good to me.

StommePoes:
Latest update was merging with v0 after chai was removed and expect.js added to replace it.
@gayanKumara the team decided they did not want the chai dependency, so you can do expects using expect.js instead of chai. The rules are very similar.

gayankumara:
@StommePoes , thanks for update , ( we did in LocaleUtil-test.js has expect for assertion.)  
so LAX team will update according to that and then separate PR will come ... 
@StommePoes  any thoughts?

gayankumara:
@StommePoes , now we could remove 'chai' and implemented unit tests with 'expect',
seems all tests passed.

StommePoes:
Yeah, and Jason already changed your calendar unit tests, so Lax only has to worry about future unit tests.

StommePoes:
Stupid stupid default Android 4.4 browser was somehow puking on my use of ES6 Sets. Seriously? I feel we should have let that browser DROWN than fixing it this way.  But so anyway, this branch should now be testable on fake-Google-Nexus @umahaea @gayanKumara.

PR #22, opened 2016-07-13
fix: remove dependencies for event harness

JasonLantz:


wyseguyonline:
What is our policy on approving PRs that lower coverage percentage?

StommePoes:
There's currently no policy that I know of...

kaadams4:
We should discuss as a team as there should be a policy! Maybe if it lowers
it a certain percentage then it should be rejected? The emails I have been
getting show a few that lowered it very small amount which seems okay.

Sent from my iPhone

On Jul 18, 2016, at 10:56 AM, Mallory <notifications@github.com> wrote:

There's currently no policy that I know of...

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/Pearson-Higher-Ed/compounds/pull/22#issuecomment-233406293>,
or mute the thread

StommePoes:
@kaadams4 A single test specific to calendar was removed, and may get replaced.

wyseguyonline:
FWIW - I am ok with percentages > 85%.  So for this PR, I'm ok with moving forward.  But we should talk about internal policies for going down on percentage (perhaps a warning rather than a fail).

umahaea:
@JasonLantz - looks good to me.

JasonLantz:
thank you Mallory.

wyseguyonline:
Looks ok to me too.

PR #23, opened 2016-07-21
feat(DayPicker): date picker calendar 

gayankumara:
fix(DayPicker) : done changes for accessibillity

fix: added click outside dependancy

### DIFF
--- a/docs/src/app/components/containers/components/DatePicker.js
+++ b/docs/src/app/components/containers/components/DatePicker.js
@@ -1,0 +1,113 @@
+import React from 'react';
+import DayPicker from './DayPicker';
+import DateUtils from './DateUtils';
+import moment from 'moment';
+import Onclickout from 'react-onclickout';
+import '../../../../www/scss/calendar.scss';
+
+export default class DatePicker extends React.Component {
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      value: '',
+      month: new Date(),
+      showComponent: false,
+      selectedDays: null
+    };
+
+    this.handleOnClickOutside = this.handleOnClickOutside.bind(this);
+    this.handleOnClick = this.handleOnClick.bind(this);
+    this.handleDayClick = this.handleDayClick.bind(this);
+    this.handleInputChange = this.handleInputChange.bind(this);
+    this.showCurrentDate = this.showCurrentDate.bind(this);
+  }
+
+  handleOnClick() {
+    this.setState({
+      showComponent: true
+    });
+
+    this.refs.input.select();
+  }
+
+  handleOnClickOutside() {
+    this.setState({
+      showComponent: false
+    });
+  }
+
+
+  handleDayClick(e, day, { selected }) {
+
+    this.refs.input.select();
+
+    this.setState({
+      selectedDay: selected ? null : day,
+      moment: moment.locale(this.props.locale),
+      value: moment(day).format('L'),
+      month: day,
+      showComponent: false
+    });
+  }
+
+  showCurrentDate() {
+    this.refs.daypicker.showMonth(this.state.month);
+  }
+
+ handleInputChange(e) {
+   const { value } = e.target;
+
+   // Change the current month only if the value entered by the user
+   // is a valid date, according to the `L` format
+   if (moment(value, 'L', true).isValid()) {
+     this.setState({
+       month: moment(value, 'L').toDate(),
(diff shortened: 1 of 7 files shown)


docs/src/app/components/containers/components/DatePicker.js
```diff
@@ -0,0 +1,112 @@
...
+      position: 'absolute',
+      backgroundColor:'white',
+      marginTop:'5px'
+    };
+
+    const divStyle1 = {
+      display: 'inline-block'
+    };
+
+    return <div className="pe-input"><label>Choose a date</label>
```
StommePoes: <label>Choose a date</label> <-- this needs a "for" attribute pointing to the id of the input.

If you had this and it didn't appear, remember that in JavaScript, "for" is a special word so for HTML it has to be called "HTMLFor".

And, because a page may have unknown number of datepickers, I think we should switch to something that can generate a unique id per datepicker-- and give that id to the input, and to the label as HTMLFor.

docs/src/app/components/containers/components/DatePicker.js
```diff
@@ -0,0 +1,112 @@
...
+          <div className="pe-input-group pe-input-icon">
+          <div className="pe-input" style={divStyle1}>
+           <Onclickout onClickOut={this.handleOnClickOutside}>
+           <input ref="input" type="text" id="i2" placeholder="mm/dd/yyyy"
+           onClick={this.handleOnClick}  value={this.state.value}
+           onChange={this.handleInputChange}
+           onFocus={this.handleOnClick} />
+
+           <span className="pe-btn pe-btn--medium" onClick={this.handleOnClick} onFocus={()=>{this.refs.input.select()}}>
```
StommePoes: You actually shouldn't need to add any click handlers to the span-- in the CSS, I've set 
```
pointer-events: none;
```
to the span-- that means if someone tries to click on the span, it will act as if it is not there and the user is really clicking on the input, and be caught by your event listeners on the input.

So you can act like the span is decoration :)

docs/src/app/components/containers/components/DatePicker.js
```diff
@@ -0,0 +1,113 @@
+import React from 'react';
+import DayPicker from './DayPicker';
+import DateUtils from './DateUtils';
+import moment from 'moment';
+import Onclickout from 'react-onclickout';
+import '../../../../www/scss/calendar.scss';
```
StommePoes: @wyseguyonline @JasonLantz I suggested this component needs custom CSS as we don't know yet if my PR for Elements will be accepted by management/the team. Is this the right way to do it?

wyseguyonline: Appears to be.  As long as they're loading their .scss files correctly (see component-archetype/webpack.config.js).

docs/src/app/components/containers/components/DatePicker.js
```diff
@@ -0,0 +1,113 @@
...
+    return <div className="pe-input"><label htmlFor="i2">Choose a date</label>
+          <div className="pe-input-group pe-input-icon">
+          <div className="pe-input" style={divStyle1}>
+           <Onclickout onClickOut={this.handleOnClickOutside}>
+           <input ref="input" type="text" id="i2" placeholder="mm/dd/yyyy"
+           onClick={this.handleOnClick}  value={this.state.value}
+           onChange={this.handleInputChange}
+           onFocus={this.handleOnClick} />
+
+           <span className="pe-btn pe-btn--medium" onFocus={()=>{this.refs.input.select()}}>
```
StommePoes: @gayanKumara you still shouldn't need this onFocus, spans normally cannot receive focus anyway (users won't be able to tab to it).

docs/src/app/components/containers/components/DatePicker.js
```diff
@@ -0,0 +1,113 @@
...
+           <i className="pe-icon--calendar" aria-hidden="true"></i>
+           <span className="pe-sr-only"></span></span>
+              {this.state.showComponent ? <div style={divStyle} >
+                <DayPicker enableOutsideDays id="i2"
```
StommePoes: Since the input tag has the id of "i2", that id has to be unique on the web page, so DayPicker will need something different.

StommePoes:
Looks good to me. Anyone else?

umahaea:
looks good to me.

PR #25, opened 2016-08-04
feat(datepickerrange): date picker range component

gayankumara:
feat: start and end date focus w/o Selection

feat: date format change according to the locale

fix: date format change and fix lint errors

feat: once change any date on right date picker then day set to changed date

fix: fixed issue with invalid locale pass through url

fix: gap issue with two date pickers

fix: focus on inputs when select day

fix:

Date Picker Range Input Part1 - If user selects Outside date from a Start Date Picker,Dates after
the Selected start date are Disabled state in End Date Date picker.

feat: when from date picker click then apprear from and to from second date picker

fix: previously selected start date is not cleared when user selects different start date

fix: user can select an end date without selecting a start date

fix: to date picker clear when click a date grater then on selected from date

fix: avoid null if cleck on same day twise

fix: issue with from less than  day

fix: issue with date box color

fix: if click on from date grater then to date on first date picker

### DIFF
--- a/docs/src/app/components/containers/components/DatePickerRange.js
+++ b/docs/src/app/components/containers/components/DatePickerRange.js
@@ -1,0 +1,245 @@
+import React from 'react';
+import DayPicker from './DayPicker';
+import DateUtils from './DateUtils';
+import moment from 'moment';
+import Onclickout from 'react-onclickout';
+
+export default class DatePickerRange extends React.Component {
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      from: '',
+      to: '',
+      month: new Date()
+    };
+
+    this.handleOnClickOutside = this.handleOnClickOutside.bind(this);
+    this.handleOnClickFrom = this.handleOnClickFrom.bind(this);
+    this.handleOnClickTo = this.handleOnClickTo.bind(this);
+    this.handledoubleClick = this.handledoubleClick.bind(this);
+    this.handleDayClickFrom = this.handleDayClickFrom.bind(this);
+    this.handleDayClickTo = this.handleDayClickTo.bind(this);
+    this.handleInputChangeFrom = this.handleInputChangeFrom.bind(this);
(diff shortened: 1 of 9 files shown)


docs/src/app/components/containers/components/DatePickerRange.js
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  handleDayClickTo(e, day) {
+    const { from, to } = this.state;
+
+    this.refs.inputTo.select();
+
+    let range = DateUtils.addDayToRange(day, this.state);
+
+    if (from === '' && (to === '' || to !== '')) {
```
StommePoes: I'm not sure what's being checked after the && really... isn't this going to always be true?

docs/src/app/components/containers/components/DayPicker.js
```diff
@@ -208,11 +208,21 @@ export default class DayPicker extends React.Component {
...
     e.persist();
-    if (modifiers.indexOf("outside") > -1) {
-      this.handleOutsideDayPress(day);
+
+    let className = "pe-calendar-dates";
+    className = modifiers.map(modifier => ` ${className}--${modifier}`).join("");
+
+    if (className !==  ` pe-calendar-dates--disabled`) {
+
+      if (className !==  ` pe-calendar-dates--outside pe-calendar-dates--disabled pe-calendar-dates--disabled`)
```
StommePoes: Do we actually get td's with two identical classes (pe-calendar-dates--disabled)? I don't think that should be valid HMTL, plus we only care if the classes contain the disabled one.

gayankumara: Thanks @StommePoes , I just done code re factor accordingly your feedback , so needs to re -test and let you know to merge it once the testing is done, (I did not re base since if something fail then need to be reverted)

meanwhile you can just go through it.

PR #26, opened 2016-08-15
feat(datepickerrange): date picker range with new user stories

gayankumara:
… branch

style: added month title for next previous

fix: lint errors fixed

fix: if user click from date if exists then to date also change as from date

fix: if user click on from date after or previous current month then to date picker display correct

feat: if user click from date grater than to on from date picker

feat: if user select from date when already selected then never goes to the to date picker

### DIFF
--- a/docs/src/app/components/containers/components/DatePickerRange.js
+++ b/docs/src/app/components/containers/components/DatePickerRange.js
@@ -1,0 +1,260 @@
+import React from 'react';
+import DayPicker from './DayPicker';
+import DateUtils from './DateUtils';
+import moment from 'moment';
+import Onclickout from 'react-onclickout';
+
+export default class DatePickerRange extends React.Component {
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      from: '',
+      to: '',
+      month: new Date()
+    };
+
+    this.handleOnClickOutside = this.handleOnClickOutside.bind(this);
+    this.handleOnClickFrom = this.handleOnClickFrom.bind(this);
+    this.handleOnClickTo = this.handleOnClickTo.bind(this);
+    this.handledoubleClick = this.handledoubleClick.bind(this);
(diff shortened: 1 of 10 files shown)


gayankumara:
Reopened since Travis CI fails....

StommePoes:
I simply re-ran Travis again. It seems to have passed. All the errors we see in the logs of the failed build are also in v0.

docs/src/app/components/containers/components/DatePickerRange.js
```diff
@@ -0,0 +1,260 @@
...
+      position: 'absolute',
+      backgroundColor:'white',
+      marginTop:'5px'
+    };
+    const divStyle1 = {
+      display: 'inline-block',
+      paddingRight:'5px'
+    };
+
+    return <div>
```
wyseguyonline: Incredibly minor nitpick, but the returned HTML should be in parenthesis.

StommePoes: Ah yeah, in fact I thought they were required for anything multiline... guess not.

wyseguyonline:
I'm not seeing any issues.

JasonLantz:
i don't see any either...

just to clarify though....

return() in the react render method lets you use multiple lines without the first line of jsx on the same line as the return.

ie. return(
           <div>
                <p>hi there</p>
          </div>
     )

if there's only one line of jsx you can write

return <input blah="blah">

JasonLantz:
screen shot 2016-08-18 at 2 46 08 pm

StommePoes:
@jasonlantz apparently React is okay with not using the () as Gayan had had no errors...

anyway, agreed the only problems I see are outside this particular PR.

There is a larger issue with Isura's related PR where datepicker and datepickerrange are added to index.js, but we'll need to track that down. I think this particular PR can be merged though.

PR #29, opened 2016-09-06
fix: remove Dropdowns test error; moved Calendar and (renamed) DatePi…

StommePoes:
…ckerCalendar to /src

What is not yet finished:
- even changing the root in .istanbul.yml doesn't remove the parsing errors you see when you run `npm test`
- I don't fully understand how code is supposed to sit in the demos, and I'm pretty certain the sole complex widget, the DropDownMenu, is not doing it right due to weirdness when inspecting react

react tools showing a <_class> element being added around the DropDown elements


StommePoes:
Ok, this is open for review, both for what I did (moving the stuff) and the 2 things I'm not sure how to fix (esprima parse errors mysteriously only on the newish calendar-related things and the demo codes).

docs/src/app/components/containers/components/Calendar.js
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
+import { Calendar } from 'pearson-compounds';
+import Demo from '../../Demo';
+import ApiDocs from '../../ApiDocs';
+
+
+const code = `
```
wyseguyonline: Can this be eliminated?

wyseguyonline:
Code looks ok to me.

StommePoes:
Good first step. Next in importance I think is finding the cause of the parsing issue. Since in master merely changing the root in .istanbul.yml fixed bugs, I think Jason is correct that it's somehow a path problem. However I'm stumped at why this is only for calendar and friends (happened even when there were no subdirectories for the helper funcs).

PR #30, opened 2016-10-26
chore: refactor compounds

JasonLantz:
normalize infrastructure with component-archtype....
go to npm_scripts for build infrastructure.  
kill multiple webpack files...

∆ main.js to an interface. renamed to Compounds.

∆ demo....
make demo single page multi section
rip out react router. 
smooth out dom instantiation, and move to demo.js
move translations to demo 
set up interface for demo components 
implement new demos

added node 7 to travis.


demo/demoCompounds/DemoButton.js
```diff
@@ -0,0 +1,32 @@
+import React      from 'react';
+import { Button } from '../../Compounds';
+
+
+export const DemoButton = (props) => {
+  return(
+    <div className="displaySection">
+      <h1 className="displayTitle"><a className="displayTitleLink" href="https://pearson-higher-ed.github.io/design/c/buttons/">Button</a></h1>
+      <div className="elementDescription">
+        <iframe src="https://pearson-higher-ed.github.io/design/c/buttons/" style={{width:'100%', height:'100%'}}/>
```
StommePoes: iframe elements are required to have a title attribute, for accessiblity. It's like the only place a title attribute is actually useful for people :P 
```
<iframe title="The Pearson Elements Design Spec page" (or something) etc>
```

demo/demoCompounds/DemoLabel.js
```diff
@@ -0,0 +1,28 @@
...
+        <p className="code">{"<Label>This is the default label.</Label>"}</p>
+        <Label type="secondary">This is a secondary label.</Label>
+        <p className="code">{"<Label type='secondary'>This is a secondary label.</Label>"}</p>
+        <Label size="small">This is a small label.</Label>
+        <p className="code">{"<Label size='small'>This is a small label.</Label>"}</p>
+        <Label size="large">This is a large label.</Label>
+        <p className="code">{"<Label size='large'>This is a large label.</Label>"}</p>
+        <Label inverse style={{ backgroundColor: '#000' }}>This is an inverse label.</Label>
+        <p className="code">{"<Label inverse style={{ backgroundColor: '#000' }}>This is an inverse label.</Label>"}</p>
+        <Label><h4>A label may include a single child element, which will be used as the root element.</h4></Label>
```
StommePoes: Since the point of label as far as Elements is concerned is to style things, are you deliberately using an h4 in there to show "look, now it won't look like a heading"?

.travis.yml
```diff
@@ -1,15 +1,23 @@
...
+    secure: I7MhEd++LqQyxQcivH7thVwazRHmV1By4TzjdKicHHur4qHaQzjiq5a5pM7LtPzrDbgxZapBwA/9okz13iSGeqG+/Nks8Nc2ac1wGOpnewnOSNnCw4e8tJT7Ug88s5cdcIfKeA0GhzUrRie5xEADBk1+9T+UXUujGiUNYcVftmqwV2xEHeE2FqjLGat0OLLx33llpM7ppt3pC00W7dRz/1IEY1Ov/U4lSqEqlppGfbLhi3X7SrgK66eqAiBS21lHUc2EF+/9Z5cebYjVMB6Wlbb6DcbYsm2rUEXiNmKu23P89yOBgLx9VEImDnyfOAfwGXX0H8fHgDXXC0dQboC9pCcBMbe6v4PH9zg0qs2pfUbsBTEX8m376dI1jm8RZhwVAQ0o1wsqLe0d11+p9hEROp0vs4Vgi7Uern7Fzyvpsi/b/ZpxIxbldqbhyb9GVgcrepfBvG3OFCVxNTcjYjyUjn1BATHZA9MdFQgsZTf//AjsbuQp6KGEPzCUTRLlzc5TvJ6LwsPw5Kvbj5hTLUblcgbPNTQH31vcHxkqWaQ8cnwSFinjiKJtdYn5SCHHpMLxdGSg2Pr16mNP+uByf6/A7tpb304TOJGzv3WQdZrwaOTG0KFRIGb4Lurjn+sZXzIsKkboUsJDijdnuxOLCl5a7aPT0JfP+I2IR0GHcAdpG0M=
   skip_cleanup: true
   on:
-    branch: v0
+    branch: master
```
aaronkaka: v0 is the master branch for the SDKs

.travis.yml
```diff
@@ -1,15 +1,23 @@
 language: node_js
 sudo: false
 node_js:
-- '6.3'
+- 4
+- 6
+- 7
```
aaronkaka: No Pearson consumers will be on Node 7 for a while. Console is the most advanced, and they will be moving to Node 6 for Bitesize. Takes more time to successfully build across 3 Node versions.

StommePoes: That and moving to 7 will need a lot of cleanup? according to all the scary warning messages.

wyseguyonline:
Can we remove the "under construction" from the demo site?

StommePoes:
@wyseguyonline only after we can release. That's our birthday treat

.travis.yml
```diff
@@ -16,7 +15,7 @@ deploy:
     secure: I7MhEd++LqQyxQcivH7thVwazRHmV1By4TzjdKicHHur4qHaQzjiq5a5pM7LtPzrDbgxZapBwA/9okz13iSGeqG+/Nks8Nc2ac1wGOpnewnOSNnCw4e8tJT7Ug88s5cdcIfKeA0GhzUrRie5xEADBk1+9T+UXUujGiUNYcVftmqwV2xEHeE2FqjLGat0OLLx33llpM7ppt3pC00W7dRz/1IEY1Ov/U4lSqEqlppGfbLhi3X7SrgK66eqAiBS21lHUc2EF+/9Z5cebYjVMB6Wlbb6DcbYsm2rUEXiNmKu23P89yOBgLx9VEImDnyfOAfwGXX0H8fHgDXXC0dQboC9pCcBMbe6v4PH9zg0qs2pfUbsBTEX8m376dI1jm8RZhwVAQ0o1wsqLe0d11+p9hEROp0vs4Vgi7Uern7Fzyvpsi/b/ZpxIxbldqbhyb9GVgcrepfBvG3OFCVxNTcjYjyUjn1BATHZA9MdFQgsZTf//AjsbuQp6KGEPzCUTRLlzc5TvJ6LwsPw5Kvbj5hTLUblcgbPNTQH31vcHxkqWaQ8cnwSFinjiKJtdYn5SCHHpMLxdGSg2Pr16mNP+uByf6/A7tpb304TOJGzv3WQdZrwaOTG0KFRIGb4Lurjn+sZXzIsKkboUsJDijdnuxOLCl5a7aPT0JfP+I2IR0GHcAdpG0M=
   skip_cleanup: true
   on:
-    branch: master
+    branch: v0
```
StommePoes: wait, there were a few other places that were changed to reference "master" instead of v0 though, right? In some scripts or something...? the shell script?

StommePoes:
Ok I've been running this and having 3 different scrolling boxes on a page is maddening.

I can make it more sane by making the following change:
```.elementContainer {
    background-color: #d3d3d3;
    padding: 10px;
}
```

That is, removing the height and width. Let the content fill that up, instead of forcing the browser to mess with overflow scrollbars.  If this was needed for a very small or very large screen then I would say a media query should limit the sizing to those viewports.

StommePoes:
@wyseguyonline I would let Parker remove "under construction" only *after* the menu worked with keyboard :P

StommePoes:
@JasonLantz I would also not reference the inputs for the LABEL component. This is another example of developer confusion due to crappy names, and Parker has been working hard on trying to fix this.

The LABEL component should be something that is part of an INPUT component. There's pretty much no reason in HTML to ever have a &lt;label/&gt; tag without an &lt;input/&gt; that it belongs to. The original LABEL component was part of Typography, and all the "pe-label-blah-blah" classes aren't for labels but for UI textual typography thingies.

I think we should remove LABEL before releasing Compounds, maybe replacing it with the ICONS instead?

StommePoes:
We're also getting errors in the console when running the demo-- something (and I cannot find what) is calling something called `https://cdnjs.cloudflare.com/ajax/libs/velocity/1.2.3/velocity.ui.min.jss` which is 403 mostly because of the typo in the name. However, why we are calling this library, what it does, do we need it? and how it is coming in with a typo, I have no idea. 

If we don't know what it's doing, can we get rid of it?

The other error is coming from an Urchin tracking script. Why should anyone be tracking me running a local demo on my localhost? Can we get rid of that? Do we really need to spy on employees???

demo/index.html
```diff
@@ -0,0 +1,25 @@
+<!-- <!doctype html> -->
```
StommePoes: Why is the doctype commented out?

README.md
```diff
@@ -41,46 +42,51 @@ After cloning the repository:
 
     npm install
     npm start
-    
-Navigate to **http://localhost:8000/#/components**, where the spawned Node server hosts a webpack-generated SPA using React Router for defining how to render the components.
+
+Navigate to **http://localhost:8000/#/components**, where the spawned Node server hosts a webpack-generated SPA using
```
StommePoes: It says here localhost:8000 however in `npm_scripts/dev.js`  we have
```
exec(`npm run dev-setup`);                                                      
exec(`webpack-dev-server --port 8081 --devtool source-map --hot --progress --colors`);
```

PR #31, opened 2016-11-09
Des 497

StommePoes:
Icons


JasonLantz:
looks good Mallory.....

wyseguyonline:
LGTM

StommePoes:
Only thing I'm unhappy with in this push is:
- gotta add in the propType checker again for buttons
- need to find a way for "readonly" props of className to be editable by the user (so they can have their own classNames *in addition to* our Elements classnames.

package.json
```diff
@@ -1,94 +1,71 @@
...
     "release": "node ./npm_scripts/release.js",
-    "compile": "babel ./src --out-dir ./lib --source-maps both",
-    "test-local": "babel-node ./node_modules/istanbul/lib/cli.js cover _mocha -- ./test",
-    "istanbul": "npm run compile && babel-node ./node_modules/istanbul/lib/cli.js cover _mocha -- ./test",
-    "coveralls": "cat coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js",
-    "test-ci": "npm run istanbul && npm run coveralls && rm -rf coverage",
-    "test": "npm run istanbul",
-    "version": "npm run gen-changelog && git add CHANGELOG.md"
+    "verify": "node ./npm_scripts/verify.js",
+    "postinstall": "cp ./node_modules/@pearson-components/npm-scripts/npm_scripts/*.js ./npm_scripts"
```
wyseguyonline: Please rename postinstall to copy-utils and change remaining script names to what is in the archetype when appropriate.

package.json
```diff
@@ -1,94 +1,71 @@
...
-    "react-router": "^1.0.0",
-    "sass-loader": "^3.1.2",
-    "semver": "^5.1.0",
-    "sinon": "^1.17.3",
-    "style-loader": "^0.13.0",
-    "surge": "^0.17.4",
-    "webpack": "^1.12.2",
-    "webpack-dev-server": "^1.12.1",
-    "moment":"2.12.0",
+    "babel-core": "^6.18.0",
```
wyseguyonline: Add @pearson-components/npm-scripts dependency.

package.json
StommePoes: does compounds now fully follow the archetype?

(that would be nice, less stuff that's different)

wyseguyonline: Follows it enough that we're using the npm-scripts stuff.

npm_scripts/release.js
```diff
@@ -16,20 +18,25 @@ const syncRemote = (branchName, nextVersion) => {
...
 };
 const exitFailure = (message) => {
-  console.error(message);
+  log.primaryError(message);
   process.exit(1);
 };
 
-if (!branchName.startsWith('v')) {
-  exitFailure('You must be on the official version branch in order to release to npm.');
+if (branchName !== 'master') {
```
StommePoes: I actually dunno why this is here. Are we going to change the branch name eventually to "master" before we ever release Compounds?

src/Icon.js
```diff
@@ -14,6 +14,7 @@ const Icon = (props) => {
            xmlnsXlink = "http://www.w3.org/1999/xlink"
            role = "img"
            aria-labelledby = {i_id}
+           focusable = "no"
```
wyseguyonline: This won't impact someone trying to make an icon tab accessible right?

StommePoes: No, this is an IE-only supported attribute. It was part of TinySVG 1.2 but never approved, IE team had gotten ahead of itself.

Making SVGs themselves focusable (outside of IE) is actually a huge pain in the ass and not easy (most people end up placing actual links inside them instead), however SVG2 plans on just supporting the usual tabindex attribute like everyone else.

This does not impact SVGs inside focusables like buttons, except removing an extra, unexpected tab stop. I'm 99% certain the reason Bob has an open bug on his help button is because they autofocus on the form and a single shift-tab in all browsers but IE brings you to the help button, so then you can hit spacebar/enter to engage it. In IE you're actually on the SVG and need a second shift-tab to reach the actual button, making it seem like the button is teh b0rked.

JasonLantz:
I pulled these changes into the ∆ pr....  if everything looks right we can scrap this pr and just work off of v0.

PR #32, opened 2017-01-03
Des 510

JasonLantz:
this installs jest as the test runner and code coverage...

run `npm test`


StommePoes:
should the copy-utils line be
```
"copy-utils": "cp ./node_modules/@pearson-components/npm-scripts/npm_scripts/*.js ./npm_scripts"
```
or is what I have outdated?

wyseguyonline:
@StommePoes 

it should be:

"copy-utils": "node ./node_modules/@pearson-components/npm-scripts/npm_scripts/cross-platform.js"

demo/translations/fr.json
```diff
@@ -0,0 +1,4 @@
+{
```
wyseguyonline: Do we need this?

JasonLantz: i may need it for proofing the intl...

PR #34, opened 2017-02-02
Des 550

JasonLantz:
event harness, Button, Icon, TextInput, PasswordInput, tests, docs.....


demo/demo.scss
```diff
@@ -108,13 +108,13 @@ li {
...
 
 //=========Button Page===============================
-svg {
-  margin-left: 10px;
-}
-
-button svg {
-  margin: 4px 0 0;
-}
+//svg {
```
wyseguyonline: Is this to be removed or just commented out for now (i.e. we will be reinstating this code at a later date)?

npm_scripts/exec.js
```diff
@@ -1,11 +0,0 @@
-// ES6 Modules support still in progress for the V8 engine, hence using 'require' here
```
wyseguyonline: Are these npm_scripts files being replaced by the ones in our npm-scripts repo?

src/Icon.js
```diff
@@ -13,6 +13,7 @@ const Icon = (props) => {
            xmlns = "http://www.w3.org/2000/svg"
            xmlnsXlink = "http://www.w3.org/1999/xlink"
            role = "img"
+           focusable = "no"
```
wyseguyonline: Should focusable be configurable?

demo/demo.scss
StommePoes: Only if it's still needed to "center" the svg in the button. There's no standard set of CSS I could find and the icon-in-a-button was a demo use case.

src/Icon.js
StommePoes: No. The icons themselves should never be focusable. IE is following the TinySVG 1.2 spec which was never actually approved, so no current (nor future) SVG spec allows for this attribute (which in IE is defaulted to true/yes).

wyseguyonline: Thanks for the clarification.

PR #38, opened 2017-03-07
Des 551

JasonLantz:
add gh pages


build/demo.compounds.js
```diff
@@ -0,0 +1,4 @@
...,c.default.createElement("div",{className:"elementContainer"},c.default.createElement("h5",null,"Single Line"),c.default.createElement(s.TextInput,{id:"a",label:"First Name",type:"default",placeholder:"First Name"}),c.default.createElement("p",{className:"code"},'<TextInput id="a" label="First Name" type="default" placeholder="First Name" />'),c.default.createElement("br",null),c.default.createElement("br",null),c.default.createElement("h5",null,"Single Line - Error"),c.default.createElement(s.TextInput,{id:"b",label:"First Name",type:"error",placeholder:"First Name"}),c.default.createElement("p",{className:"code"},'<TextInput id="b" label="First Name" type="error" placeholder="First Name" />'),c.default.createElement("br",null),c.default.createElement("br",null),c.default.createElement("h5",null,"Single Line - Disabled"),c.default.createElement(s.TextInput,{id:"c",label:"First Name",type:"disabled",placeholder:"First Name"}),c.default.createElement("p",{className:"code"},'<PasswordInput id="a" label="Password" />'),c.default.createElement("br",null),c.default.createElement("br",null),c.default.createElement("h5",null,"Single Line - Read Only"),c.default.createElement(s.TextInput,{
```
wyseguyonline: Can we .gitignore these files?

JasonLantz: added to gitignore....

node_modules/.bin/eslint
```diff
@@ -0,0 +1 @@
+../eslint/bin/eslint.js
```
wyseguyonline: We shouldn't have anything from node_modules in this repo either.

README.md
```diff
@@ -10,7 +10,19 @@ This project supports [Node v4+](https://nodejs.org) and npm 2+ installed in you
...
+    ```
     npm install pearson-compounds --save
+    ```
+
+    Import a single component:
+    ```
+    import { Button } from 'pearson-compounds';
+    ```
+
+    Import a all components:
```
StommePoes: import a all?

demo/demoPages/ButtonPage.js
```diff
@@ -1,52 +1,64 @@
...
+      this.state = {};
+    }
+
+    render() {
+      return (
+      <div className="displaySection">
+        <h2><a href="https://pearson-higher-ed.github.io/design/c/buttons/">Button</a></h2>
+
+        <div className="elementContainer">
+          <h5>Icon Button:</h5>
```
StommePoes: let's not skip heading levels. If we want that text to look like an h5, I think we have a class for that. High-school essay teachers will faint when they see this :P

demo/demoPages/ButtonPage.js
```diff
@@ -1,52 +1,64 @@
...
+          <Button btnSize="large">Large</Button>
+          <p className="code">{'<Button btnSize="large">Large</Button>'}</p>
+          <br />
+          <h5>XLarge Button:</h5>
+          <Button btnSize="xlarge">XLarge</Button>
+          <p className="code">{'<Button btnSize="xlarge">XLarge</Button>'}</p>
+          <br />
+          <h5>Primary XLarge Button:</h5>
+          <Button btnType="primary" btnSize="xlarge">Primary XLarge</Button>
+          <p className="code">{'<Button btnType="primary" btnSize="xlarge">XLarge</Button>'}</p>
```
StommePoes: so as a user, I can 
```
<Button btnType="primary" className="foobar">foo</Button>
```
?

demo/demoPages/HeaderPage.js
```diff
@@ -0,0 +1,43 @@
...
+      this.state = {};
+
+      this.toggleList = _toggleList.bind(this);
+  }
+
+  render() {
+    return  (
+      <div className="container">
+        <div className="header">
+          <Link to="/"><h4 className="headerTitle">Pearson Design Accelerator</h4></Link>
```
StommePoes: why h4?

demo/demoPages/HomePage.js
```diff
@@ -0,0 +1,44 @@
...
+    super(props);
+    this.state = {};
+  }
+
+
+  render() {
+    return (
+        <div className="homepage">
+
+          <h4>Atomic React Components</h4>
```
StommePoes: why h4

demo/demoPages/HomePage.js
```diff
@@ -0,0 +1,44 @@
...
+          <ul className="homepageList">
+            <li><Link className="link_homepage" to={`buttons`}>Button</Link></li>
+            <li><Link className="link_homepage" to={`icons`}>Icon</Link></li>
+            <li><Link className="link_homepage" to={`inputs`}>Inputs</Link></li>
+          </ul>
+
+          <p>Contributions: <a href="https://github.com/Pearson-Higher-Ed/docs/blob/master/origami-contributions.md">FAQ</a></p>
+          <p>Pearson Elements SDK: <a href="http://pearson-higher-ed.github.io/elements/">http://pearson-higher-ed.github.io/elements/</a></p>
+          <p>Design Site: <a href="http://pearson-higher-ed.github.io/design/">http://pearson-higher-ed.github.io/design/</a></p>
+          <p>Unit Test Coverage: <a href="http://localhost:8081/coverage/lcov-report/">http://localhost:8081/coverage/lcov-report/</a></p>
```
StommePoes: is this supposed to point to a live URL, or is it meant that devs have to locally run to see test coverage?

demo/demoPages/HomePage.js
```diff
@@ -0,0 +1,44 @@
...
+          <p>HipChat Room: PDA-Help</p>
```
StommePoes: Are we later linking to this?

demo/demoPages/InputsPage.js
```diff
@@ -1,98 +1,98 @@
...
+        <h2><a href="https://pearson-higher-ed.github.io/design/c/inputs/beta">Inputs</a></h2>
+
+        <div className="elementContainer">
+
+
+          <h5>Single Line</h5>
+          <TextInput
+            id="a"
+            label="First Name"
+            type="default"
```
StommePoes: Like with buttons, we really can't get away with using 
```
type=
```
as a React prop because it will be interpreted by browsers as a valid property with an invalid value. I suggest inputType or something.

demo/demoPages/InputsPage.js
```diff
@@ -1,98 +1,98 @@
...
+            placeholder="First Name"
+          />
+          <p className="code">{'<TextInput id="a" label="First Name" type="default" placeholder="First Name" />'}</p>
```
StommePoes: If we're adding the label as a proptype of the TextInput, is this where we will add in hints, error messages etc as well later?

demo/demoPages/InputsPage.js
```diff
@@ -1,98 +1,98 @@
...
+
+
+          <h5>Single Line - Disabled</h5>
+          <TextInput
+            id="c"
+            label="First Name"
+            type="disabled"
+            placeholder="First Name"
+          />
+          <p className="code">{'<PasswordInput id="a" label="Password" />'}</p>
```
StommePoes: Remember, developers copy code. We need unique id's (and a unique-id-generator in the component) for users (also for the error messages and anything else that will need aria-whatever sticking to it). 

I have already seen from the Polar Bear team our SVG remove-18 icon repeated multiple times on a single page with the same id and the same &lt;title&gt; text) :/

StommePoes:
```
loaderUtils.parseQuery() received a non-string value which can be problematic, see https://github.com/webpack/loader-utils/issues/56
parseQuery() will be replaced with getOptions() in the next major version of loader-utils.
Hash: 5cf56f649ccedbfb0767
Version: webpack 2.2.1
Time: 9953ms
```
is this something we need to deal with?

StommePoes:
Other than all my comments above, and after figuring out I needed https in the URL and need to override my browser's security warnings, this looks good.

I think the DES-607 to add in the newer icons will also add in a sane top-margin for SVGs inside buttons of all the sizes so they vertically center.

demo/demoPages/HomePage.js
JasonLantz: i don't what the this is.... Hipchat?

demo/demoPages/InputsPage.js
JasonLantz: i don't think that's how it works... the type will always be a prop passed down to the input. it would have to be associated with the type attribute to cause problems. also, we were breaking out the input type, TextInput, emailinput, passwordInput.... But, i'll change it to avoid confusion...

demo/demoPages/InputsPage.js
JasonLantz: ideally props would be the way of passing error messages down to each component.

demo/demoPages/InputsPage.js
JasonLantz: good point with the generator.... lemme chew on that for a sec

JasonLantz:
i believe the error your referencing will be addressed with the next update on the styles module....

Sorry about the https... the inputs show as insecure on gh_pages if gh_pages isn't enforcing https... so i set the local server to reflect the env we're deploying the docs to...

demo/demoPages/InputsPage.js
StommePoes: I only remember the buttons (but this is way back when with different code) had type=default, type-primary getting exposed directly to the HTML so that's why I had changed it for buttons.

If they don't get exposes straight anymore then we're cool.

demo/demoPages/HomePage.js
StommePoes: I think so. Do we want a link to the room here? (how public is that room, is it for outside Pearson consumers too or....?)

umahaea:
There is nothing to QA in here. Since the regression tests pass, we should be good to merge this unless there is any other comment.

StommePoes:
I agree, but I have a question: can your inputs tests you're finishing up run on this branch? As in, the moment 551 is merged, and your tests are done, in theory those new tests could run against master?

umahaea:
Yes mallory inputs tests I have been writing it against the "v0" branch.

PR #44, opened 2017-04-18
added icons to demo page; added prop for pe-icon--btn class

StommePoes:
... I still can't figure out how someone could use one of our buttons AND propTypes AND add their own classes though. React is frustrating.

### DIFF
--- a/src/Icon.js
+++ b/src/Icon.js
@@ -5,11 +5,9 @@ import uuid                 from 'uuid';
 const Icon = (props) => {
 
   const icon_class = 'pe-icon--' + props.name;
-  let i_id;
+  const i_id = '_'+uuid.v1();
 
   if (props.children) {
-    i_id = '_'+uuid.v1();
-
     return(
       <svg role = "img"
            focusable = "false"
--- a/src/Icon.js
+++ b/src/Icon.js
@@ -10,10 +10,7 @@ const Icon = (props) => {
     const i_id = '_'+uuid.v1();
 
     return(
-      <svg version = "1.1"
-           xmlns = "http://www.w3.org/2000/svg"
-           xmlnsXlink = "http://www.w3.org/1999/xlink"
-           role = "img"
+      <svg role = "img"
            focusable = "false"
            aria-labelledby = {i_id}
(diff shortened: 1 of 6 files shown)


JasonLantz:
well... there's the though that they shouldn't be able to add their own classes... and the props control the applied state of the component...

StommePoes:
I don't see at all how it will be possible for any team to use these without adding their own classes. I have yet to see another team use these *without* adding their own classes. 

Besides, it's not a problem in regular Javascript so I'm a little miffed it's so damn hard in React. I hate frameworks.

src/Icon.js
```diff
@@ -10,10 +10,7 @@ const Icon = (props) => {
     const i_id = '_'+uuid.v1();
```
wyseguyonline: Why declare and initialize here rather than at the top of the function?

StommePoes: where do you want it? outside the if statement?

wyseguyonline: In order to avoid confusion and possibly introducing bugs due to hoisting, variables should be declared (and with const, initialized) at the top of the scope, in this case the function.

StommePoes: I thought vars were hoisted. And can I
const thing = ""
then later if props 
const thing = uuid

Because I thought const won't let me do that.

If I need an empty var then I will switch to let.

src/Icon.js
```diff
@@ -5,15 +5,13 @@ import uuid                 from 'uuid';
 const Icon = (props) => {
 
   const icon_class = 'pe-icon--' + props.name;
+  let i_id;
```
wyseguyonline: Unless there is a reason why we can't/shouldn't run uuid.v1() here, let's just do

const i_id = '_' + uuid.v1();

StommePoes: It's only needed when we need unique id's, which is only when the icon has a title tag inside it for the aria-stuff.

StommePoes: But I don't care if you want it all the time I'll do that.

wyseguyonline: Since it is a const, let's move it up outside the if.

PR #45, opened 2017-04-19
Des 611 - Dropdown component

mhomolak:


src/Dropdown.js
```diff
@@ -0,0 +1,151 @@
...
+                           <span className="mobile-title-text">
+                             {this.props.mobileTitle}
+                           </span>
+                           <button className="mobile-close-button">
+                             <svg
+                               id="header-close"
+                               aria-hidden="true"
+                               focusable="false"
+                               className="pe-icon--remove-lg-18">
+                               <use xlinkHref="#remove-lg-18">Selected</use>
```
StommePoes: This text is
- loose inside an SVG without meaning
- going to be maybe interpreted as the button's name

The button being the control is the thing that needs a name, ideally for all the cases where we see any kind of labelly text, it would be also the name (as in accessible name) of the button.

Additionally, if we're going to have any accessible-names inside a control purely by using the SVG, we need to follow our specs for SVG (or use the <Icon/> component which, so long as you give it both the class and some text "inside" it, will do all that for you).

Most importantly we don't want a dropdown to accidentally have control names like "selected" esp since the developers don't always see this text.

src/Dropdown.js
```diff
@@ -0,0 +1,151 @@
...
+                                          className="li-props">
+                                        <button type="button" id="mobile-font" className="li-button">
+                                        { this.props.presentationType !== 'label'
+                                          ?
+                                           <svg
+                                             id={`svg-id${appendId}`}
+                                             aria-hidden="true"
+                                             focusable="false"
+                                             className="pe-icon--check-sm-18">
+                                             <use xlinkHref="#check-sm-18">Selected</use>
```
StommePoes: again here we need to decide:
is the svg just a pretty picture and needs aria-hidden?

if not, and it needs to have a text equivalent, we need to use the &lt;title&gt; tag and remove aria-hidden (this acts like the "alt text" of the svg image).  To use as a component, do &lt;Icon name="check-sm-18"&gt;Selected&lt;/Icon&gt;.

src/Dropdown.js
```diff
@@ -0,0 +1,151 @@
...
+  render() {
+
+    const appendButtonClass = containerMargin.marginRight > 0 ? '-right' : '';
+
+    const ddLabel = this.props.presentationType === 'label';
+    const ddButton = this.props.presentationType === 'button';
+    const ddIcon = this.props.presentationType === 'icon';
+
+    return(
+        <div onClick={this.toggleDropDown} ref={(div) => { this.dropdown = div; }}
```
StommePoes: ... maybe for later, but do we want a pattern of click listeners on divs, or do we want these on already-clickables and style these buttons to match designs so we get all the buttony stuff for free?

src/Dropdown.js
```diff
@@ -0,0 +1,151 @@
...
+             className="dropdown-container" style={containerMargin}>
+          { ddLabel ? <div>
+                        <p className="dropdown-label-text">{this.props.presentationText}</p>
+                        <button className={`icon-btn${appendButtonClass}`}>
```
StommePoes: So this works... there's just some loose text first, then a button with the accessible name of "open".

src/Dropdown.js
```diff
@@ -0,0 +1,151 @@
...
+      selectedItem: ''
+    }
+
+    this.toggleDropDown = _toggleDropDown.bind(this);
+    this.selectedItem = _selectedItem.bind(this);
+  }
+
+  renderListItems() {
+    let items = [];
+    const mobileHeader = <li className="mobile-header" key="header">
```
StommePoes: I think I figured out why this bothers me. It reminds me of something Bootstrap used to do: they'd make a "fake header" from one of the list items, even though semantically it was nothing more than a list item.

src/Dropdown_scss/Dropdown_fullView.scss
```diff
@@ -0,0 +1,101 @@
...
+}
+
+.li-wrapper {
+  border: 1px solid #c7c7c7;
+  width: 100%;
+  padding: 12px 0;
+  margin-top: 2px;
+  margin-bottom: 0;
+  border-radius: 2px;
+  display: inline-block;
```
StommePoes: you're probably going to want to make this position: absolute so that opening and closing the dropdown doesn't shift the page underneath it.

test/Dropdown.test.js
```diff
@@ -0,0 +1,81 @@
...
+      this.wrapper = shallow(<Dropdown presentationType="icon"
+                                       list={items}
+                                       mobileTitle="Title" />);
+      expect(this.wrapper.find('button').node.props.children.props.name).toEqual('dropdown-open-sm-24');
+    });
+
+    it('has correct state', function () {
+      const document = jsdom('');
+      const window = document.defaultView;
+      Object.keys(document.defaultView).forEach((property) => {
```
StommePoes: why do we make a const window that = doc defaultview if we never use it?

src/Dropdown_scss/Dropdown_mobile.scss
```diff
@@ -13,11 +18,14 @@
...
     top: 0;
     left: 0;
+    right: 0;
+    bottom: 0;
     background-color: #FFFFFF;
     position: absolute;
     margin-top: 0;
     padding: 0;
     height: 100%;
+    overflow: scroll;
```
StommePoes: you can also, if you only want scrolling in 1 direction, use overflow-y: scroll too.

src/Dropdown.js
```diff
@@ -69,7 +70,7 @@ class Dropdown extends Component {
                                              <use xlinkHref="#check-sm-18"></use>
                                            </svg>
                                           : null }
-                                            <span className="dropdown-item">{item}</span>
+                                            <span className="dropdown-item" role="menuitem">{item}</span>
```
StommePoes: It should go on the interactive element (the button), not the span. Otherwise the browser sees this:
```
<presentation/no-role element>
<button (role=button)>
<menuitem/>
</button>
</presentation>
```
and according to the spec, a menu (item with a role of menu, the ul) must have menuitems (or menuitemcheckboxes or menuitemradios) as the direct children. This is why we put "presentation" on the li elements, because they have a natural role of listitem. 

So it's easiest to move the menuitem role on the button, but if there is a reason not to do that, then the button needs a role=presentation added to it.

mhomolak:
Got it, I had misunderstood.  I'll change that.

StommePoes:
I just ran this locally:
the buttons used to open/close the dropdowns have no text.

The icon-only button should use either aria-label="name of button" or the &lt;Icon&gt; with the text inside it so there's an accessible name inside the SVG.

StommePoes:
The buttons inside the list all have the same id.

I *strongly* recommend we use something like uuid to create unique id's. 

Afterwards, the rendered page should be run through an HTML validator or something to catch these.

udhadpa:
Tests written and verified. Good to merge

PR #47, opened 2017-05-01
Des update elements

StommePoes:


src/Button.js
```diff
@@ -3,17 +3,19 @@ import React, { PropTypes } from 'react';
...
+  }
+  else if (!btnType) {
     classes = (!btnSize) ? `pe-btn` : `pe-btn--btn_${btnSize}`;
-  };
-
+  }
   if (btnType) {
     classes = (!btnSize) ? `pe-btn__${btnType}` : `pe-btn__${btnType}--btn_${btnSize}`;
-  };
+  }
```
wyseguyonline: I think we should clean this up a bit.  Would the following work?

```
let classes = (!btnSize) ? `pe-btn` : `pe-btn--btn_${btnSize}`;
if (btnIcon) {
  classes = `pe-icon--btn`;
}
if (btnType) {
  classes = (!btnSize) ? `pe-btn__${btnType}` : `pe-btn__${btnType}--btn_${btnSize}`;
}
```

StommePoes: Would be great if, in addition, other classes could be added to the buttons. Otherwise, we can use this basic button component in, for example, the modal, where a class is used to float it to the right. We don't want people injecting inline styles just because they can't add classes.

wyseguyonline: Agreed.  However, looking at how the classes are structured, they're not additive.  I think we might need a story to refactor how these classes are applied.  Would allow us to write more simple branching logic and allow users to more precisely define what their buttons will look like without having to inject their own styles.

For example, in this case we apply class names like pe-btn--btn_${some button size} and pe-btn__${some type}--btn_${some button size} which are probably amalgams of a number of styles that could be split out.  I might recommend breaking them down into several classes each of which strictly governs one aspect of the button's appearance.  For example  pe-btn (defines general button styles applicable to all Pearson butons), pe-btn-${sizeName} (defines the size), pe-btn-${buttonType} (defines the type of button), then any additional classes (e.g. a class to make the button an icon) that might be necessary.  Then our branching logic could be just a series of if statements that gradually adds more specificity to the button's appearance by adding class names to the string that is eventually dropped into the markup.

StommePoes: That's what I would like to do: a base style that covers 90%, then additional classes for things like size and type. 
I did try earlier to first read the button's className before overriding it with classes but I think I'm doing the reading-of-props things wrong, plus since we're writing them they might not make sense  to be props

StommePoes: Also, we can prolly remove the small size entirely. It seems to exist solely to be definitive when we mean a default-sized button. However so long as you don't add large or xlarge or whatever, you get a small button.

umahaea:
Regressions passed. Good to merge.
Desktop: https://travis-ci.org/Pearson-Higher-Ed/ux-test-platform/builds/234897826
Mobile: https://travis-ci.org/Pearson-Higher-Ed/ux-test-platform/builds/234902397

umahaea:
merging it.

PR #50, opened 2017-05-15
Des 534 - Tabs component

mhomolak:
With the exception of the style tweaks still coming from design (font stuff), the component is ready for review.


StommePoes:
Notes:

1. There's aria-hidden="true" on the currently-selected item. That makes things invisible. You meant `aria-current="true"`... I may have promised some docs on that, my bad. https://tink.uk/using-the-aria-current-attribute/
But you definitely-definitely-definitely want to get rid of all the aria-hidden stuff! You also don't need aria-current="false" on the other tabs unless that's easier; there can also simply be no attribute at all to mean false as well.

2. Since you're not hiding and showing from a group of pre-existing content divs, you'll probably have to make the div with the inline padding a live region (or the div around it, if that inline-padding-div is content the consuming team feeds in...).
Give whichever div we have control over `aria-live="polite"` to let screen reader users know there's new content. Unfortunately this will cause all of whatever in that "pane" to start reading out which can be annoying, but users can start navigating into that page to get back to their own pace.

3. During the Dev & Design meeting, Design stated this would primarily be used for navigation at the top of a page, and thus a list of links that brought users to new pages. They stated they would come up with some visually-different design for actual "tab-panels" where small chunks of content change under the "tabs." Because users tab through a list of links, and those links have href's that actually Go Somewhere, while a tab panel requires users to switch to arrow keys and clicking a "tab" uses no href and merely changes a small chunk of content underneath, I'm in a bit of a pickle with this component as it's currently coded.

This code uses href="#", which if used as a navigation menu as Design claimed, we'd need to accept an array of URLs, which it currently doesn't, so now we have links that don't Go Somewhere and that's kinda not great (in tab-panel with roles, the links aren't links anymore semantically because they get role="tab" on them). In general, wherever we have something you click that Does Something, with an href that goes nowhere, which means semantically it should be a button, not a link.

It doesn't help that, because of Google not knowing semantics of HTML, material-design hangers-on are calling lists of links (typical website navigation menus as we've known since 1995) "tabs" when there's already a separate established navigation pattern called a "tab panel" which consists of "tabs" and "panes." 

I dunno that I can okay this as-is from a governance perspective. I wonder if we need to bring this up with Design again, and find out exactly where Console is using this today.

mhomolak:
Closing for now

PR #51, opened 2017-05-18
chore: merge in v0

JasonLantz:


StommePoes:
We do need some changes.
The first one, you did say we lost the wrapper. I'm not sure if that was also doing the aria-hidden, but so on Linux and Windows screen readers, when a modal is open, I still have access to the rest of the page. The tabs are trapped in the modal (good) but the rest of the page is still available. I can for instance still go to the next or previous heading, link, whatever item.
I'm not sure the best route to fix this now; in the demo we could put aria-hidden="true" on the div id="app" but not sure what most teams have. But app is holding most of the page except for the actual modals so that's kinda what we want.

2. When I'm focussed on the close button or any other buttons, I can't see that I'm focussed on them. As a sighted keyboarder I'm lost.  The first modal without the footer, the default-style button is darker, but since design made the hover/focus styles so subtle, it's hard to tell that by looking at it. Firefox will still have a focus ring in there because it'll do that for buttons no matter what, but other browsers suffer.

3. like in my notes to Eajaz:
```
<div class="ReactModal__Content ReactModal__Content--after-open pe-template__static-medium modalContent" 
tabindex="-1" role="dialog" aria-label="Modal"
aria-labelledby="modalContent">
```
In the future we may move focus to the dialog itself, but we're not doing that now, so we don't need the tabindex="-1" at all.
You can also remove the aria-labelledby (because the id that it points to, modalContent, doesn't seem to exist, and also because aria-label (which does use a string for its value) overrides all names for all the things.
Either remove the aria-labelledby OR remove the aria-label (users can't change that string through props, right? so removing aria-label might be better) and make aria-labelledby point to the id of the title heading ("modalHeaderText"). If you do that, when the dialog opens most SR users should hear "dialog [the text inside the title/h2]" which lets the consuming teams choose their modal name basically.

4.  &lt;button **type=button** (suggestion only-- the default type is a submit and if somehow some team manages to wrap our stuff in a form tag I wouldn't want things accidentally submitted)
** aria-expanded** (set to "true" when modal is open and "false" when it's not.
class="modalClose pe-icon--btn"&gt;

5.  &lt;svg version="1.1" *xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"* <-- we can dump these, they're in the sprite now, \o/
 focusable="false" role="img" aria-hidden="false" aria-labelledby="_740520e0-3bfc-11e7-8685-9d2b5ff4d04f" class="pe-icon--remove-sm-24"&gt;&lt;title id="_740520e0-3bfc-11e7-8685-9d2b5ff4d04f"&gt;close dialog&lt;/title&gt;&lt;use 
*xmlns:xlink="http://www.w3.org/1999/xlink"* <-- same, we can dump this one
 xlink:href="#remove-sm-24"&gt;&lt;/use&gt;&lt;/svg&lt;

I was going to say get rid of the tabindex on the body div but I see that we never know when that part might need to scroll, so... leave it in there, but because it's hard to tell when I can scroll in it, I'd say let's leave the outline visible. Normally I do hide it for scrollable divs but here, I notice not seeing it means I have trouble knowing when my arrows will work.

6. After the modals are dismissed, their manually-added padding they have on their parents remains behind. So the demo with 3 modals, after opening all three, there's a big empty area at the bottom of the screen.

JasonLantz:
so, couple of things...

1) I did have a version with a 1 px dotted border that was like the browser defalut focus ring.... should i put that in on the focused element?

2) i'll have to look at putting the wrapper back... now we have the ability to pull in multiple modals on the page with es6 context.  so, import {modal as myModal} from "pearson-compounds".... i'll have to look at how to we're excluding the element.

3) I think you're looking at eajaz's test config.... the modalContent id is there when the modal is shown..... if you think it better to pull it, I will...

4) Mike is doing the update to the buttons... should we mention this to him....

5) k, pulling the xmlns.... the use tag looks like this.... <use xlinkHref={'#' + name}></use>

6) opening pr on elements removing the outline:none; on the modalBody...  this pr can add the focus ring in the modal mentioned above.... as for the padding... good catch. i've got this handled for the change i added for hung yesterday, but the event isn't triggering the js needed to reset the style... i've made it trigger the style removal js.

StommePoes:
>I did have a version with a 1 px dotted border that was like the browser defalut focus ring.... should i put that in on the focused element?

Or just remove the outline: none and let whatever the browser does through.

>i'll have to look at putting the wrapper back... now we have the ability to pull in multiple modals on the page with es6 context. so, import {modal as myModal} from "pearson-compounds".... i'll have to look at how to we're excluding the element.

Thanks, it's a similar problem some other components have actually. We've even bounced the idea around of a dedicated component who does nothing but group the rest-of-page into an element we can aria-hide it.

>I think you're looking at eajaz's test config.... the modalContent id is there when the modal is shown..... if you think it better to pull it, I will...
The id is on the whole modal though. That means everything inside it would be "labelling" the modal. Instead, the id of the title (h2) works.
Eajaz' config didn't have it, right, but at the last minute I did find it on yours, but that id isn't a good place for labelling in this case.

>Mike is doing the update to the buttons... should we mention this to him....
Yeah, but also removing outline: none would at least give some default style.
I kinda wish design came up with some awesome focus styles, funky borders or something. That would be fun.

>k, pulling the xmlns.... the use tag looks like this.... <use xlinkHref={'#' + name}>

++

>opening pr on elements removing the outline:none; on the modalBody... this pr can add the focus ring in the modal mentioned above.... as for the padding... good catch. i've got this handled for the change i added for hung yesterday, but the event isn't triggering the js needed to reset the style... i've made it trigger the style removal js.

cool

umahaea:
Good to merge.

umahaea:
sorry not this one..

StommePoes:
namm them icons :P

PR #52, opened 2017-05-18
fix: Dropdown styles moved back to Compounds

mhomolak:


udhadpa:
Regression passed. Good to merge

PR #53, opened 2017-05-24
Des 534 - Tabs component

mhomolak:


src/Footer/Footer.js
```diff
@@ -0,0 +1,82 @@
...
+
+    this.state = {
+      position: ''
+    }
+  }
+
+  renderCopy() {
+    const year = new Date().getFullYear();
+    return (<p>
+              Copyright &copy; {year} Pearson Education Inc.
```
wyseguyonline: We should leave the "Copyright &copy; {year} " but since this is an open source component available to the vast interwebz, should we make the "Pearson Education Inc." part configurable?

JasonLantz:
there shouldn't be any hard coded text in the component... pass it from translations in the demo...

src/Footer/Footer.js
mhomolak: Good call.  I'll add in another prop and use that with Jason's suggestion too.

mhomolak:
I just realized I must have munged branches so this got some of the Footer commits.  I pushed the updates to the actual Footer branch DES-471.

demo/demoPages/TabsPage.js
```diff
@@ -0,0 +1,73 @@
...
+          <div style={eight}>{`</Pane>`} <br/> </div>
+          <div style={eight}>{`<Pane label="Tabberino">`} <br/> </div>
+          <div style={sixteen}>{`<div>Tab 3</div>`} <br/> </div>
+          <div style={eight}>{`</Pane>`} <br/> </div>
+          {`</Tabs>`}
+        </p>
+
+      <div className="code">
+        <h2>Tab Props</h2>
+        <h4>Required:</h4>
```
StommePoes: Lawlz. We shouldn't get in the habit of skipping heading levels in docs 'cause we KNOW other teams will do it too.

src/Tabs/Pane.js
```diff
@@ -0,0 +1,26 @@
...
+
+  componentDidUpdate() {
+    if (this.props.onActive) this.props.onActive();
+  }
+
+  render() {
+    const i_id = uuid.v1();
+
+    return (
+      <div role="tabpanel" id={i_id}>
```
StommePoes: since uuid may generate something starting with a digit, and since that's no-no for HTML id tokens, add a _ or something in front of it. _ is a legal starter

src/Tabs/Tabs.js
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  componentDidMount() {
+    const parentUl = this.doc;
+    const tabArray = parentUl.querySelectorAll('[role=tab]');
+    const lastTabArray = tabArray.length - 1;
+
+    parentUl.addEventListener("keydown", (event) => {
+      const selectedIndex = this.state.selected;
+      if (selectedIndex === 0 && event.keyCode === 37) {
```
StommePoes: Not sure but this might be able to be shortened if the main if (or case) is the e.which and then it's either next-or-first, prev-or-last.

QA should check if keyCode works in all modern browsers. I know e.which is the same everywhere. Old browsers used to have issue with keyCode but maybe none of the modern ones do.

webpack.config.js
```diff
@@ -7,6 +7,9 @@ const main              = `${__dirname}/demo/main.js`;
 const compounds         = `${__dirname}/index.js`;
 const icons             = `${__dirname}/node_modules/pearson-elements/dist/icons/p-icons-sprite-1.1.svg`;
 const elements          = `${__dirname}/node_modules/pearson-elements/dist/css/elements.css`;
+const Footer            = `${__dirname}/src/Footer/Footer.scss`;
+const Tabs              = `${__dirname}/src/Tabs/Tabs.scss`;
```
mhomolak: I'm not sure why that's still showing but I don't have those const's locally.  Might be leftover from when I accidentally munged the Footer & Tabs branches.

src/Tabs/Tabs.js
```diff
@@ -0,0 +1,111 @@
...
+
+  }
+
+  renderLabels() {
+    function labels(child, i) {
+      let activeClass = this.state.selected === i ? 'activeTab' : '';
+      let tabI = activeClass ? "0" : "-1";
+      let ariaSelected = activeClass ? true : false;
+      const themeCheck = this.props.light ? 'light' : '';
+      let i_id = uuid.v1();
```
StommePoes: or let i_id = '_'+uuid.v1();

StommePoes:
All of the tabs have the same ID. uuid needs to be reworked to:
1- on load make all the tabs, give each one their own unique id
2- on tab click the only thing the click handler should change for them is their active class, tabindex, and aria-selected
3- the active div/panel's aria-describedby points to the active tab's id. It's a bit weird that only one pane exists at any given time so the aria-describedby needs to update to match active tab id I guess.

src/Tabs/Tabs.js
```diff
@@ -0,0 +1,111 @@
...
+      }
+    }, true)
+
+  }
+
+  renderLabels() {
+    function labels(child, i) {
+      let activeClass = this.state.selected === i ? 'activeTab' : '';
+      let tabI = activeClass ? "0" : "-1";
+      let ariaSelected = activeClass ? true : false;
```
StommePoes: So, the tabindex and the ariaSelected parts probably should go up to the didMount func (since it runs at least once onload anyway it should work) and then this whole func should only ever run once... maybe in WillMount? so clicking a tab should never call this renderLabels again since they only ever need 1 rendering.

src/Tabs/Tabs.js
```diff
@@ -0,0 +1,111 @@
...
+      const themeCheck = this.props.light ? 'light' : '';
+      let i_id = '_'+uuid.v1();
+
+      return (
+        <li key={i} role="presentation">
+          <a href="#"
+             role="tab"
+             id={i_id}
```
StommePoes: A simple way to make these unique per tab could be to add the key as a string, like
id={i_id + '_${key}`}
(or something, I prolly have syntax wrong).

I believe when using uuid for the old Laxapana calendar, calling uuid directly also ensured a unique id per item:
id='_'+uuid.v1()

Then I wonder if we can copy that prop of id to something in state, so when we renderContent() we can pass it as a prop to the current pane??

PR #60, opened 2017-06-15
Des 685 - Banner style alert

mhomolak:


JasonLantz:
run a: git rm -r build/ 
and commit....

everything else looks fine...

StommePoes:
Only thing is, when I run the demo, it's being added after all the other stuff on the page.

I can see any headers or app-header coming first, but this alert should be inserted as a first-child (or whatever) of the app element. Meaning on a page with a lot of focusables, my tab should end up on the closing x quite soon (again, possibly excepting headers since visually it's slightly below the top of the page).

PR #62, opened 2017-06-22
Des 707 - Functionality updates for Console

mhomolak:
- Two new alignment props allowing the consumer to have the dropdown menu "dropup" or be right aligned.
- When the dropdown menu is open, any click outside of the menu will now cause it to close.


demo/demo.scss
```diff
@@ -65,7 +65,7 @@ h2 a {
   height:50px;
 }
 
-.header button {
+.header .headerButton {
   float:right;
   margin:2.5px;
```
StommePoes: Are we okay with browsers rounding this to 3px or 2px? Apparently some of them round these while others don't.

src/Dropdown/Dropdown.js
```diff
@@ -83,28 +102,30 @@ class Dropdown extends Component {
...
-          { ddButton ? <div className={`pe-btn-container${appendButtonClass}`}>
+          { presentationType === 'button' ?
+                       <div className={`pe-btn-container${rightAlign}`}>
                          <div className="pe-btn__primary">
-                           {this.props.presentationText}
-                           <button className={`icon-btn${appendButtonClass}`}>
+                           {presentationText}
+                           <button className="icon-btn">
                              <svg
                                id="icon-in-button"
```
StommePoes: So, question: if Console has a dropdown for each course in a list of courses and they call this Dropdown component for each one... does that mean each one will have an icon on the single web page with this exact same ID?

demo/demo.scss
JasonLantz: is it important?  It is in the demo....

StommePoes: No idea. Are we using it for Console anywhere?

JasonLantz: demo scss isn't in the dist bundle, so it shouldn't affect the end user....

src/Dropdown/Dropdown.js
mhomolak: That id is left over from doing something before.  I can remove it.

demo/demo.scss
StommePoes: ok

PR #65, opened 2017-06-27
Console dropdown

mhomolak:
Console is using React router to inject links & a button into the Dropdown.

- Added prop to disable tabIndex
- Fixed an error being caused in their app due to each Dropdown having an eventListener.
- Escape will now close the Dropdown menu when open.

### DIFF
--- a/demo/demoPages/DropdownPage.js
+++ b/demo/demoPages/DropdownPage.js
@@ -71,7 +71,6 @@ const DropdownPage = () => (
 
           <h3>Icon only (selectable options)</h3>
             <Dropdown
-              killFocus
               presentationType="icon"
               list={listItems}
               mobileTitle={mobileTitle}
--- a/demo/demoPages/DropdownPage.js
+++ b/demo/demoPages/DropdownPage.js
@@ -71,6 +71,7 @@ const DropdownPage = () => (
 
           <h3>Icon only (selectable options)</h3>
             <Dropdown
+              killFocus
               presentationType="icon"
               list={listItems}
               mobileTitle={mobileTitle}
--- a/demo/demoPages/DropdownPage.js
+++ b/demo/demoPages/DropdownPage.js
@@ -32,6 +32,8 @@ const DropdownPage = () => (
             <li className="li-props">alignRight:Boolean === {`<Dropdown alignRight />`}</li>
             <li className="li-props">By default, a dropdown menu is left aligned. Add the&nbsp;
               <i>alignRight</i> prop to right align the <br/> dropdown menu.</li>
+            <li className="li-props">killFocus:Boolean === {`<Dropdown killFocus />`}</li>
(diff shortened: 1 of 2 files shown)


src/Dropdown/Dropdown.js
```diff
@@ -161,7 +167,8 @@ export default class Dropdown extends Component {
 
 };
 
-function _toggleDropDown(e) {
+function _toggleDropdown(e) {
```
JasonLantz: you don't need the e....

demo/demoPages/DropdownPage.js
```diff
@@ -16,16 +16,16 @@ const DropdownPage = () => (
...
-            <li className="li-props">A different view is used for mobile and this title will be reflected
-              in the mobile header <br/>when the dropdown is open.</li>
             <li className="li-props">list:Array = ["one", "two", "three"] || ["one", "two", "divider", "three"]</li>
             <li className="li-props">The "divider" option may be used in conjunction with any of
             the presentationType for <br/>separating options presented in your list.</li>
           </ul>
           <h3>Optional:</h3>
           <ul>
             <li className="li-props">presentationText:String === "Your label or button text"</li>
+            <li className="li-props">mobileTitle:String === "Your mobile title"</li>
```
JasonLantz: should the killFocus prop be listed here?

mhomolak: I wasn't sure.  Do we think this will be needed in the future?

mhomolak: I guess if it's an option I might as well add it.

wyseguyonline: If it is an option, we need to document it.  I'm not keen on hiding features that only one team will use since it seems counterproductive in the long term.

mhomolak: It's added in the newer commits.

src/Dropdown/Dropdown.js
```diff
@@ -49,6 +54,12 @@ export default class Dropdown extends Component {
     }
   }
 
+  handleEsc(e) {
+    if (e.which === 27) {
+      this.setState({ open: false });
+    }
```
StommePoes: focus needs to be placed back on the dropdown opening button if we close via esc, esp if user is currently focussed on a submenu item, that will vanish from the DOM.

src/Dropdown/Dropdown.js
```diff
@@ -71,6 +82,8 @@ export default class Dropdown extends Component {
       const appendId = this.state.selectedItem === item
                        ? '-this.state.selectedItem' :'';
 
+      const killFocus = this.props.killFocus ? '-1' :'0';
+
       const dividerLine = <li className="divider-container" key={i}>
```
StommePoes: this li by default I believe still has its list-item role, no? There's a role for separators though, this could be given a role="separator"

udhadpa:

PR #68, opened 2017-07-05
fix: remove listener binding

nlkluth:
bound click handlers are treated as separate callbacks (til), so the `removeEventListener` wasn't working correctly; this meant `handleOutsideClick` was still firing after the component was unmounted, causing uncaught exceptions


mhomolak:
Cool, I also did not know that!

mhomolak:
One node version passed.  Cancelling the other builds to merge and trigger a release.

PR #70, opened 2017-07-07
fix: esc key once again closes the DD menu

mhomolak:
After the recent change that bound this with handleOutsideClick, handleEsc also needed to be bound for the proper context in setting state.


udhadpa:
Regression passed for Dropdown Tests - https://travis-ci.org/Pearson-Higher-Ed/ux-test-platform/builds/251251385
Good to merge.

PR #71, opened 2017-07-17
Des 540 (For Mallory)

mhomolak:


src/Calendar/Calendar.js
```diff
@@ -0,0 +1,162 @@
...
+    }
+  }
+
+  handleKeys = (e) => {
+    if (e.which === 37 || e.which === 38 || e.which === 39 || e.which === 40 || e.which === 13) {
+      e.preventDefault();
+    }
+
+    switch (e.which) {
+      case 13: console.log(e.which); break; // enter
```
wyseguyonline: Are we console logging every keypress for a reason or is this just placeholder functionality for now?

mhomolak:
Placeholder for actual keys setup.  This is just a PR to give Mallory a
look at what was already done. I'll close it after she get's a look.

src/Calendar/Calendar.js
```diff
@@ -0,0 +1,162 @@
...
+      selectedYear: date.getFullYear(),
+      selectedMonth: date.getMonth(),
+      selectedDate: date.getDate(),
+      selectedDt: new Date(date.getFullYear(), date.getMonth(), date.getDate()),
+      startDay: 0,
+      minDate: this.props.minDate ? this.props.minDate : null,
+      disablePast: this.props.disablePast ? this.props.disablePast : false,
+      dayNames: ["S", "M", "T", "W", "T", "F", "S"],
+      monthNamesFull: ["January", "February", "March", "April", "May", "June",
+        "July", "August", "September", "October", "November", "December"],
```
StommePoes: These are moving elsewhere eventually, right? Because they'll be determined by translations and thus can't be hardcoded part of the component?
We may instead need hardcoded ... codes, which later in the component can call from translations and stuff the right string in.

mhomolak:
Yeah they will need to be moved elsewhere.

src/Calendar/Calendar.js
```diff
@@ -0,0 +1,162 @@
...
+    const state = {};
+    if (this.state.month < 11) {
+      state.month = this.state.month + 1;
+      state.year = this.state.year;
+    } else {
+      state.month = 0;
+      state.year = this.state.year + 1;
+    }
+    Object.assign(state, this.calc.call(null, state.year, state.month));
+    this.setState(state);
```
StommePoes: So, am I remembering wrong, or do we want to set only the specific properties of state that we're changing (instead of updating the whole state each time)? Or am I mis-reading this?

src/Calendar/Calendar.js
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  render() {
+    const { monthNamesFull, month, year, dayNames, startDay, daysInMonth,
+            firstOfMonth, selectedDate, disablePast, minDate
+          } = this.state;
+
+    return (
+      <div className="r-calendar">
+        <div className="r-inner">
```
StommePoes: what's the r stand for? Weren't we prefixing everything with "pe" so people knew they were PDA/UXF component classes?

mhomolak: row, it can be changed

src/Calendar/Calendar.scss
```diff
@@ -0,0 +1,135 @@
+.r-calendar {
+  border: 1px solid #c7c7c7;
+  background-color: #fff;
+  min-width: 320px;
+  padding: 16px 0 20px 0;
+  height: 320px;
+  width: 320px;
+  position: absolute;
```
StommePoes: I wonder if the standalone calendars in general need to be pos: abso. If it's only for pickers, maybe the absolute positioning should be moved to the pickers.

src/Calendar/Calendar.scss
```diff
@@ -0,0 +1,135 @@
...
+}
+
+.inner-cell-square {
+  height: 32px;
+  width: 32px;
+  align-items: center;
+  display: flex;
+  justify-content: center;
+
+  &:hover {
```
StommePoes: The currently-selected (as user keyboards through by arrows) will also match the :hover state. Only when they hit enter will it become the selected-date state.

deepak2510:
Hey guys,

I'm still getting notifications for this :)

src/Calendar/Calendar.js
mhomolak: Only daysInMonth, firstOfMonth, month and year are being updated.

src/Calendar/Dates.js
```diff
@@ -0,0 +1,91 @@
...
+
+  statics = {
+    year: new Date().getFullYear(),
+    month: new Date().getMonth(),
+    date: new Date().getDate(),
+    today: new Date(new Date().getFullYear(), new Date().getMonth(), new Date().getDate())
+  }
+
+  render() {
+    let haystack;
```
StommePoes: lol

src/Calendar/Header.js
```diff
@@ -0,0 +1,34 @@
...
+      <div className="r-row r-header">
+        <div className="r-cell r-title pe-title--small">
+          {this.props.monthNames[this.props.month]}
+          <span className="fake-title-space" />
+          {this.props.year}
+        </div>
+        <div className="r-cell r-icons"
+             onClick={this.props.onPrev.bind(this)}
+             role="button"
+             tabIndex="0"
```
StommePoes: okay this one I don't see why we don't just use buttons with a type=button... that's like one less attribute :P

mhomolak:
@deepak2510 I removed you as a collaborator from the repo

src/Calendar/Calendar.scss
mhomolak: I'm not seeing that in other calendars.  The "selected" styles are what rove around through keyboard control.

src/Calendar/Calendar.scss
mhomolak: This could be done.  Jason said he would add a wrapper to date picker if we want to pursue this so it could be abso positioned.

PR #72, opened 2017-07-31
Inputs fix

JasonLantz:


StommePoes:
so clean it squeaks

umahaea:
Regression passed. good from QA.

PR #74, opened 2017-08-04
Elewc 2316

loahou04:


loahou04:
@JasonLantz why would we want onKeyUp rather on keydown?  Also don't we still want keyup to be on the document level?  Though the focus does seem like it could be at the component level because they'll still be focused on the component.

JasonLantz:
could be an onKeyDown.... if you use onKeyDown on the level of the dropdown then the event listeners are scoped to the dropdown which unmounts (and takes the eventlisteners with it) with a selection from the dropdown...

loahou04:
Ya we do unmount as well on componentWillUnmount.  but i agree with you
document level events aren't great.  i can see if i change it to component
level if it will still work properly...

JasonLantz:
it's fine... just push the code.

JasonLantz:
looks like the CI is failing on your namming of drop down item....
screen shot 2017-08-04 at 2 48 40 pm

demo/demoPages/DropdownPage.js
```diff
@@ -1,105 +1,131 @@
...
-              presentationText="label"
-              list={simpleList}
-              mobileTitle="Mobile title"
-            />`}
+                dropdownControlLabel="Dropdown open"
+                mobileTitle={mobileTitle}
+                changeHandler={(item) => {
+                  console.log(item);
+                }}
+                targetElement={
```
mhomolak: Weren't we avoiding using this? line 88 too

loahou04:
yep forgot to update the documentation...

src/Dropdown/Dropdown.js
```diff
@@ -58,130 +87,142 @@ export default class Dropdown extends Component {
...
-              transitionEnterTimeout={300}
-              transitionLeaveTimeout={300}>
-             {this.renderListItems()}
-            </CSSTransitionGroup> }
+        <div className="dropdown-container" onClick={this.toggleDropdown} onBlur={this.closeDropdown} >
+          {this.insertAnchor()}
+          <ul
+            role="menu"
+            id={`${this.props.id.replace(' ', '_')}-dropdown`}
+            ref={(dom) => { this.list = dom; }}
```
mhomolak: This ref is a leftover from previous implementation.

loahou04: no it is not.  I am using the list to go through the keydown events

umahaea:
all changes tested. good to merge.

PR #76, opened 2017-08-16
chore: Links open a new browser tab

mhomolak:
Console requested that Footer links open up in their own tab after being clicked.

### DIFF
--- a/src/Footer/Footer.js
+++ b/src/Footer/Footer.js
@@ -9,12 +9,14 @@ export default class Footer extends Component {
     copyrightText: PropTypes.string,
     links: PropTypes.array.isRequired,
     light: PropTypes.bool,
-    singlePageStick: PropTypes.bool
+    singlePageStick: PropTypes.bool,
+    anchorTarget: PropTypes.oneOf(['_blank', '_self', '_parent', '_top'])
   }
 
   static defaultProps = {
     light: false,
-    singlePageStick: false
+    singlePageStick: false,
+    anchorTarget: "_self"
   }
 
   constructor(props) {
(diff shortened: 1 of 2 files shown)


src/Footer/Footer.js
```diff
@@ -34,7 +34,7 @@ export default class Footer extends Component {
     for (let i = 0; i < this.props.links.length; i++) {
       let item = this.props.links[i];
       items.push(<li key={i}>
-                   <a href={item.href}>{item.text}</a>
+                   <a href={item.href} target="_blank">{item.text}</a>
```
wyseguyonline: Should this be configurable?

mhomolak: I can make it so it is.

JasonLantz: screen shot 2017-08-16 at 4 23 57 pm

wyseguyonline: Yeah.  Let's make it a configurable option.  Will have to add it to the documentation as well and there might be a QA piece to add, but I think this will make for a better component.

mhomolak: Done

src/Footer/Footer.js
```diff
@@ -9,12 +9,14 @@ export default class Footer extends Component {
     copyrightText: PropTypes.string,
     links: PropTypes.array.isRequired,
     light: PropTypes.bool,
-    singlePageStick: PropTypes.bool
+    singlePageStick: PropTypes.bool,
+    anchorTarget: PropTypes.oneOf(['_blank', '_self', '_parent', '_top'])
```
wyseguyonline: Outstanding!

umahaea:
many approvals, good to merge.

src/Footer/Footer.js
StommePoes: I'm curious why we're even offering parent and top. These were invented specifically for framed pages, which we either don't have or if we do I will have to hunt them down, kill and eat them.

mhomolak:
They were only included to offer all of the attributes that target would
accept.

PR #77, opened 2017-08-21
Des 732

mhomolak:
Placeholder PhoneNumber component

### DIFF
--- a/demo/demoPages/PhoneNumberPage.js
+++ b/demo/demoPages/PhoneNumberPage.js
@@ -46,12 +46,12 @@ const PhoneNumberPage = () => (
       </div> <br/>
 
     <PhoneNumber
-    country="US"
-    placeholder="555-555-5555"
-    labelText="Mobile phone"
-    onChange={(num) => console.log(num)}
+      country="US"
+      placeholder="555-555-5555"
+      labelText="Mobile phone"
+      onChange={(num) => console.log(num)}
     />
-    
+
     <h3>Example usage</h3>
     <p className="code">
       {`<PhoneNumber`} <br/>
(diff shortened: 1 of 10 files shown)


demo/demoPages/PhoneNumberPage.js
```diff
@@ -0,0 +1,70 @@
...
+
+const PhoneNumberPage = () => (
+  <div className="displaySection">
+    <h1>Phone Number</h1> <br/>
+    <div elementContainer>
+      <div className="code">
+        <h3>Props</h3>
+        <h4>Optional:</h4>
+          <ul>
+            <li className="li-props">labelText:String === {`<PhoneNumber labelText="Your label" />`}</li>
```
StommePoes: should this be made required? Labels in *general* are for our inputs (or should be).

src/PhoneNumber/rrui.css
```diff
@@ -0,0 +1,335 @@
...
+/* Not introducing an animation on `.rrui__select__option`
+   because it would introduce a little clutter due to
+   focused option animation in autocomplete mode */
+.rrui__select__selected
+{
+	transition         : background-color 30ms ease-out, color 30ms ease-out;
+	-webkit-transition : background-color 30ms ease-out, color 30ms ease-out;
+}
+
+.rrui__select__selected-content
```
StommePoes: will check in IE

demo/demoPages/PhoneNumberPage.js
mhomolak: This is only going to be used as a "test" component for PI in staging.  For our actual implementation I agree, it should be required but for this I didn't set it as such.

src/PhoneNumber/rrui.css
StommePoes: Looked good in IE even with the flex.

PR #80, opened 2017-08-29
Fix: add keys

briandunnpearson:
To avoid this ReactJS message in the Console tab:  "Warning: Each child in an array or iterator should have a unique "key" prop."


mhomolak:
Merging and triggering a release.

PR #81, opened 2017-08-31
Des 541

JasonLantz:
time picker, datepicker, inputs updated with componentWillReceiveProps....


test/TimePicker.test.js
```diff
@@ -0,0 +1,120 @@
...
+
+    it('should call the changehandler', function() {
+      this.wrapper = mount(<TimePicker id="test" labelText="test" timeFormat="hh:mm" changeHandler={() => {}}/>);
+      this.wrapper.find('input').simulate('click');
+      const e = {"target":{"value":"Hi There"}};
+      this.wrapper.instance().changeHandler(e);
+      expect(this.wrapper.instance().state.timePickerValue).toEqual("Hi There");
+    });
+
+    // it('should handle the inputKeyEvents down key', function() {
```
mhomolak: Still needed?

src/TimePicker/TimePicker.js
```diff
@@ -0,0 +1,203 @@
...
+      }
+      break;
+    case 13:  //keypress: enter
+      e.preventDefault();
+      this.listHandler(e);
+      break;
+  };
+};
+
+function _applyTimePickerStyles(inputState) {
```
wyseguyonline: I wonder if we should create a tech debt item (because I don't want to delay releasing this further) should be changing this (and its corresponding function for date picker) to a method inherited from the text input class.

src/TimePicker/TimePicker.js
```diff
@@ -0,0 +1,203 @@
...
+  this.props.changeHandler.call(this, e.target.value);
+};
+
+function _listHandler(e){
+  this.setState({ timePickerValue:e.target.innerText, displayOpen:false, labelStyleTmp:this.state.labelStyle });
+};
+
+function _inputEvents(e){
+  switch(e.which){
+    case 40:  //keypress: down arrow
```
wyseguyonline: This looks like a good practice to comment on which button is being pressed.

src/TimePicker/TimePicker.js
StommePoes: That sounds good.

src/TimePicker/TimePicker.js
JasonLantz: i figure comments get pulled by webpack anyway....

demo/demoPages/DatePickerPage.js
```diff
@@ -0,0 +1,108 @@
...
+              <h2>DatePicker:</h2>
+              <h3>Props:</h3>
+
+              <ul>
+                <li>time:Boolean  === "render datepicker as timepicker"</li>
+                <li>className:String  === "styles to pass to datepicker"</li>
+                <li>id:String === "A unique name for the datepicker"</li>
+                <li>dateFormat:String === "format for date/time entry"</li>
+                <li>inputState:String === "styles for input state, one of 'error','disabled','readOnly','default'"</li>
+                <li>labelText:String === "unique lable for the input field"</li>
```
StommePoes: lable -> label

demo/demoPages/TimePickerPage.js
```diff
@@ -0,0 +1,111 @@
...
+
+        <div className="elementContainer">
+
+
+          <div className="code">
+            <h2>TimePicker:</h2>
+            <h3>Props:</h3>
+
+            <ul>
+              <li>className:String  === "styles to pass to datepicker"</li>
```
StommePoes: s/datepicker/timepicker x 2?

demo/translations/en-US.json
```diff
@@ -1,13 +1,14 @@
...
+  "hideText": "hide",
+  "isNotVisibleMsg": "Password is hidden",
+  "isVisibleMsg": "Password is visible",
+  "passwordPlaceholder": "Enter Password",
+  "passwordLabel": "Password",
+  "textLabel": "First Name",
+  "textPlaceholder": "First Name",
+  "textInputInfoMessage": "This is an info message",
+  "textInputErrorMessage": "This is an error message",
+  "textInputErrorMessage": "This is an error message",
```
StommePoes: why twice error?

demo/translations/fr.json
```diff
@@ -1,13 +1,15 @@
...
+  "shownText": "Mot de passe est caché",
+  "hiddenText": "Mot de passe est visible",
+  "passwordPlaceholder": "Entrer le mot de passe",
+  "passwordLabel": "Mot de passe",
+  "textLabel": "Prénom",
+  "textPlaceholder": "Prénom",
+  "textInputInfoMessage": "Ceci est un message d'information",
+  "textInputErrorMessage": "Ceci est un message d'erreur",
+  "textInputMessage": "Ceci est un message d'information",
+  "textInputErrorMessage": "Ceci est un message d'erreur",
```
StommePoes: same

src/TimePicker/components/List.js
```diff
@@ -0,0 +1,29 @@
...
+
+import './List.scss';
+
+
+export const List = (props) => {
+
+    const { id, itemsToList, itemToParent, selectedItem, listRef, listEvents } = props;
+
+    return (
+      <ul id="itemlist" className="pe-itemList" ref={listRef} onKeyDown={listEvents} role="listbox" aria-expanded="true" aria-live="polite">
```
StommePoes: aria-live: are we dynamically adding/removing options to the list? If not, we shouldn't need a live region on this.

StommePoes:
Datepicker: In Internet Explorer, hitting tab from an open datepicker just moves my focus to the top of the page. I can cycle endlessly from the first input to the top and back again. In both IE, Firefox, and Chrome clicking anything inside the calendar part just closes the whole thing. Can't choose a date with the mouse. Keyboard can type a date, but cannot get focus into the calendar.

Timepicker:
In IE, choosing a time with the first input brings my focus back to the top of the page. In FF and Chrome, my focus seems to sorta vanish, although with a screen reader running, the screen reader's own cursor is still in place. Hitting tab from a completed timepicker in Firefox brings visible focus to that input. In Chrome it brings me to the next input. In IE focus goes to the top.

In all three browsers I am able to choose a time with mouse and keyboard, with or without a screenreader running. However to dismiss the timepicker I must click a time-- I can't click elsewhere on the document to close it.
I suspect screen readers are a little confused since when they get to the input, it created a div that says it's open-- some screen readers think that means this is an autocomplete. Down arrow works as expected, but after choosing a time, since the list vanishes, screen readers don't announce the aria-selected time (but if focus is on the input, it will of course state the current value, so people are able to get it if they need it).

StommePoes:
I have run the date and timepickers again.

Both components still have the issue of not explicitly returning the focus to the input when the dropdown/popup/whatever closes. This leaves browsers to just randomly do whatever with the focus.

Secondly, using keyboard on the datepicker now works, however I cannot make it go away by moving focus elsewhere. Calendar remains hanging open and eventually lower calendars obscure the upper calendars. Whenever focus leaves the widget's entirety, like with dropdowns, any hanging popupy things should always close up. 

I was able to select dates and hear error messages with screen readers on Windows. IE doesn't expose the month name as I go into the month, but I hear the month name anyway as part of the aria-label of the days I move through anyway so that's fine (and I don't believe we can fix it anyway). 

As a note, JAWS calls yyyy "yiyi" and has for years. Users may need to spell the label to figure that format out, but anyway... :)

Sauravvaidya:
@wyseguyonline Issues that I have found when using this component:

1) Like Mallory said on her last comment, the only to close the date and time picker is to select the date or time, isn't the widget supposed to go away if the user clicks anywhere else or basically if the input loses focus?
2) For console, we need to allow user to pick dates that are from past as many instructors create course after start date. Date picker does not seem to allow this.
3) For console, we need a time picker which does not have a labelText and timeFormat. For a11y console right now does have a label with class="pe-sr-only"
image

StommePoes:
>For console, we need to allow user to pick dates that are from past as many instructors create course after start date. Date picker does not seem to allow this.
I thought we only have disabled past dates if we put in that prop (there's a prop for past dates).

For #3 if you need hidden labels that's something I would think any input should be okay with, but the format being only available as a placeholder is not good (I can't check my chosen typed-in time/whatever against your wished-for format because placeholders vanish). Just a note

Sauravvaidya:
@StommePoes 
I agree to your reply for #3. I will check with Parker and see if he plans to have new designs for us.
For #2 All the past dates are disabled and a user cannot pick from them.
image
Also the docs page has no information for disabled prop
image

StommePoes:
Hm you're right, and I'm wondering if we lost that (did we get rid of that prop and decide that a future release would allow past dates?) or if it's still there and simply no-longer documented... i'll ask.

StommePoes:
Alright looks like hitting the enter key opens the calendar.

1. If we're going to need a keystroke to open it (which is fine), let's use the one all the other calendars including the native HTML5 datepicker user-- alt+downarrow.

2. Also, hitting Esc does close the calendar but focus is left somewhere unknown-- should focus on the input even if I had tabbed into the calendar. Same as closing dropdowns, popups, etc.

umahaea:
all approved. good to merge.

umahaea:
https://travis-ci.org/Pearson-Higher-Ed/compounds/jobs/283813467

PR #82, opened 2017-09-05
Des 766 - React Table

mhomolak:
This needs the changes made to Elements by Mallory but besides that it is good to review.

@wyseguyonline @umahaea @udhadpa @StommePoes


demo/demo.scss
```diff
@@ -31,7 +31,7 @@ h2 a {
 .code {
   background-color: #f5f5f5;
   border: 1px solid #808080;
-  font-size: .85em;
+  font-size: .9em;
```
StommePoes: \o/

demo/demoPages/ButtonPage.js
```diff
@@ -5,7 +5,7 @@ import { Button, Icon } from '../../index';
 
 const ButtonPage = () => (
       <div className="displaySection">
-        <h2><a href="https://pearson-higher-ed.github.io/design/c/buttons/">Buttons</a></h2>
+        <h2><a href="https://pearson-higher-ed.github.io/design/c/buttons/" target="_blank">Buttons</a></h2>
```
StommePoes: :( can't we teach people how to shift-click instead? :P

demo/demoPages/TablePage.js
```diff
@@ -0,0 +1,156 @@
...
+        <li className="li-props">Scope specifies whether the cell is a header
+          for a column, row, or group of columns or rows.  Defaults to "col".
+        </li>
+        <li className="li-props">columnSort:Function</li>
+        <li className="li-props">Applies the appropriate Icons to show the column is sortable and allows a sorting Function
+          to be passed in on the data.  <br/>Toggles from 'ASC' to 'DESC'.
+        </li>
+        <li className="li-props">**The following props are only necessary on Selectable tables.**</li>
+        <li className="li-props">inputId:String</li>
+        <li className="li-props">The inputId is passed to the checkbox and its label.</li>
```
StommePoes: should maybe here mention this string can only start with letters or an underscore? Dunno how many consumers know this.

demo/demoPages/TablePage.js
```diff
@@ -0,0 +1,156 @@
...
+        <li className="li-props">**The following props are only necessary on Selectable tables.**</li>
+        <li className="li-props">inputId:String</li>
+        <li className="li-props">The inputId is passed to the checkbox and its label</li>
+        <li className="li-props">containerId:String</li>
+        <li className="li-props">This must match what was passed into the TableHeaderCell component.  It is referenced here in the aria-labelledby of the checkbox.</li>
+        <li className="li-props">labelledbyCellId:String</li>
+        <li className="li-props">labelledbyCellId must match the cellId ans is passed into the aria-labelledby of the checkbox.</li>
+        <li className="li-props">cellId:String</li>
+        <li className="li-props">cellId needs to be passed to the TableRowCell that follows the checkbox.</li>
+      </ul>
```
StommePoes: What if someone wants an actual caption inside the table (that's one option; another is just something else with the correct class before or after the table like a heading)? What if they have a column of numbers, how can they add a class to right-align etc?

src/Table/components/TableHeaderCell.js
```diff
@@ -0,0 +1,111 @@
...
+  }
+}
+
+TableHeaderCell.contextTypes = {
+  selectable: PropTypes.bool,
+  sortable: PropTypes.bool
+}
+
+function _selectAll() {
+  const checkboxes = document.querySelectorAll('div.pe-checkbox input');
```
StommePoes: This should hit all checkboxes on the webpage-- even those outside tables. We probably don't want that.

I wonder if it's better to have the state of the Master Uber checkbox in "state" and find all checkboxes belonging to THIS table and set them to match that state. 
As a separate question, if we have document.blah does this also mean we can't as easily port this to server-side?

src/Table/components/TableRow.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const TableRow = (props) => {
+  const { children } = props;
+
+  const selected = () => {
+    const tables = document.querySelectorAll('.pe-table--selectable');
```
StommePoes: While this is a good bit of code for a regular-JS website, it's code that then only one table on any page would need: instead, if someone needs 3 of these on a single page, this code gets redundant fast. Not sure what the perf results would be but we'd be storing tables and parts of them in memory multiple times for example.

Including this on every single table on a page would be silly... is there a way we can have this as a separate bit of code that gets added once to a document if a selectable table is called?

demo/demoPages/TablePage.js
mhomolak: Sure, can add.

src/Table/components/TableRow.js
mhomolak: Do we have use cases for multiple tables on a page at the moment?  If so, then yeah, having this broken out would be something to discuss.

demo/demoPages/TablePage.js
mhomolak: I'll add the alignment options.

src/Table/components/TableHeaderCell.js
mhomolak: If we want to make it more precise I was thinking possibly adding another class and then adding that to the qSA as well.  I can look at replacing the docs with refs.  I need to look into the issue more though on what else is required for our components to function both in the browser and from the server-side.

demo/demoPages/TablePage.js
StommePoes: Yeah, headings (or toolbars) are accounted for. Captions have the nice thing of being inside the table and being technically able to be displayed at the top, bottom, or even side of a table (not sure who all supports side as far as browsers but it's an old setting). 
If we don't hear a clamour for it, that's fine. Paul likes to ding tables for not having a caption but the real issue is, did the table get a "name" and that's doable with either a caption OR a heading above it-- they're not required by WCAG to actually always have captions. I did allow for it with CSS in Elements though.

Could be an optional feature for later as well.

src/Table/components/TableHeaderCell.js
StommePoes: Or if there were some way of getting a ref to the current table and then doing
tableref.querySelectorAll(...) or similar. That would at least always limit the query to THIS table.

src/Table/components/TableRow.js
StommePoes: We've got teams with dashboards with more than one table... but I'm not sure about Console specifically. Could be a tech debt thing for the future I think.

src/Table/components/TableHeaderCell.js
mhomolak: The only issue with that is the ref would need to be defined in <Table /> then passed down through context to the cell.  Not sure if that's possible.

src/Table/components/TableRow.js
StommePoes: I have no idea either. We could give the table an id, but that's ick.

demo/demoPages/TablePage.js
mhomolak: Maybe bring it up to him and see? If it's something you think we should have I can add it in.  It would just mean another couple of props.

src/Table/components/TableHeaderCell.js
```diff
@@ -0,0 +1,118 @@
...
+
+  componentWillUnmount() {
+    document.removeEventListener('keydown', this.handleKey);
+  }
+
+  componentDidUpdate() {
+    if (this.props.columnSort) this.props.columnSort();
+  }
+
+  iconToggle = () => {
```
wyseguyonline: Would rewriting this as follows make it more readable?

iconToggle = () => {
  const { iconName } = this.state;
  if (iconName === 'sort-up-18') {
    return this.setState({ iconName: 'sort-down-18' });
  }

  return this.setState({ iconName: 'sort-up-18' });
}

src/Table/components/TableHeaderCell.js
```diff
@@ -0,0 +1,118 @@
...
+    }
+  }
+
+  render() {
+    const { children, scope, inputId, containerId, inputLabel, columnSort,
+            alignCell } = this.props;
+    const { selectable, sortable } = this.context;
+    const { iconName } = this.state;
+    const sortClass = sortable ? 'pe-table__sortable' :'';
+    const columnAlignment = alignCell === 'center' ? ' pe-table__center'
```
wyseguyonline: Let's rework this.  Ternary inside a ternary is a bit confusing:

How about trying:

const columnAlignment = (alignCell === 'center' || alignCell === 'right') ? 'pe-table__' + alignCell : '';

mhomolak: Nice change, fixing it.

src/Table/components/TableHeaderCell.js
mhomolak: Done and yes I think it makes it more readable.

udhadpa:

PR #83, opened 2017-09-08
actually-selected date given as string, also secondary dates. Needs c…

StommePoes:
…leanup.

The creating classes and strings (which are hardcoded and so not internationalised) in general needs cleanup, if there's a better place to make things.

I noticed that NVDA calls every focussed day "selected" regardless of who has aria-selected. So I added text "chosen day" to make it clear which date was actually selected. 
The texts for secondary day is lost since aria-label overrides all other texts, so added this to aria-label.
The text for current-day seems not needed-- Firefox and IE, NVDA and JAWS all are supporting aria-current="date" correctly. \o/

I noticed the focus outline in Firefox is terrible for the prev/next buttons, but didn't fix.

@mhomolak @wyseguyonline @umahaea @udhadpa


umahaea:
all approved. writing the test into DES-541 branch...

PR #87, opened 2017-09-14
fix: If passed in a different start day, the aria-label follows with …

mhomolak:
…the appropriate full day name

While doing screen reader testing, Jack discovered that their French calendars, which started with Monday, were lagging behind one day on what the screen reader announced.  This fixes that.


umahaea:
Regression passed: https://travis-ci.org/Pearson-Higher-Ed/ux-test-platform/builds/275664264
Good from QA.

PR #88, opened 2017-09-15
Issue86

loahou04:


umahaea:
Kiran, pls review this fix and approve it.

StommePoes:
When something is selected (with a button so user doesn't follow a link), focus is not brought back to the triggering button. 

When opening a dropdown, tabbing away to somewhere else doesn't close it, unless the next thing I focus on is a dropdown and I open it. Dropdowns should collapse whenever they lose focus. I personally know this is a pain in the ass to do with Javascript if the dropdown is the last item on the page, but all other instances are doable.

umahaea:
all approved. so merging it.

PR #89, opened 2017-09-20
chore: insert deprecation notice into README

wyseguyonline:


README.md
```diff
@@ -1,3 +1,7 @@
+#Important Notification
```
mhomolak: Does a space need to be after the '#'? I notice it's not bolded like the others.

PR #90, opened 2017-10-11
Console changes

nlkluth:
- the `columnSort` function was getting called multiple times when inside componentDidUpdate if multiple headers had it. Updating one caused the other to rerender and call the `columnSort` again and the result was an infinite loop

- the keydown listeners on global didn't seem to be necessary and the function they were calling didn't invoke the `selectAll` function properly; seemed like a bug so I cleaned it up. Along those same lines, hitting spacebar on a focused checkbox activates `onClick` so I also removed the `onKeyDown` prop from the input itself.

- cleaned up an unused variable

- added `defaultIcon`, though console won't use this -- i started going down this path and thought it might be a good feature but I can remove it

- added `iconName` (the reason we won't need `defaultIcon`) that lets us control what icon to display - we don't allow multiple sorts so we are going to want to reset the icons when a different header is sorted

- updated demo to use state when clicking on a header (to recreate `componentDidUpdate` bug) and to use `iconName` prop as an example

### DIFF
--- a/demo/demoPages/TablePage.js
+++ b/demo/demoPages/TablePage.js
@@ -11,156 +11,188 @@ const sixteen = {paddingLeft: 16};
 const twentyFour = {paddingLeft: 24};
 const thirtyTwo = {paddingLeft: 32};
 
-const TablePage = () => (
-  <div className="displaySection">
-    <h2><a href="http://pearson-higher-ed.github.io/design/c/tables/" target="_blank">Table</a></h2>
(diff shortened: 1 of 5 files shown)


umahaea:
Will write a test for this if needed.

umahaea:
good to merge: https://travis-ci.org/Pearson-Higher-Ed/ux-test-platform/builds/288673604

PR #93, opened 2017-10-18
Calendar cleanup

mhomolak:
While I was changing the event listener yesterday I saw a few opportunities to simplify some things I had done in Calendar.

- Removed unnecessary class based child components in favor of functional components.
- Further destructured props in two of the three child components.
- Changed `dateFormat` in Datepicker to have a defaultProp of the US format to eliminate a variable and from having to include that prop from the consumers end on every instance of the component.

### DIFF
--- a/demo/demoPages/DatePickerPage.js
+++ b/demo/demoPages/DatePickerPage.js
@@ -47,7 +47,7 @@ class DatePickerPage extends Component {
                 <li>time:Boolean  === "render datepicker as timepicker"</li>
                 <li>className:String  === "styles to pass to datepicker"</li>
                 <li>id:String === "A unique name for the datepicker"</li>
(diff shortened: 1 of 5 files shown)


StommePoes:
Other than whatever causes these:
Failed tests: 
  calendarTestBeforeClass(elementsSDKTests.functionalTests.CalendarTest)
  datePickerBeforeClass(elementsSDKTests.functionalTests.DatePickerTest)

, seems good.

mhomolak:
Eajaz knows the cause and is working on a fix.

umahaea:
CI passes now: https://travis-ci.org/Pearson-Higher-Ed/compounds/builds/289732522

umahaea:
I know what is the problem with those failed tests. I will put a permanent fix for it sometime sooner.

StommePoes:
cools

PR #96, opened 2017-11-10
fix: Cleanup unused variables

mhomolak:


umahaea:
Looks good, will do a full regression run before we can approve it.

udhadpa:
Passed.

PR #98, opened 2017-11-17
Mobilefix

loahou04:


src/Dropdown/Dropdown.js
```diff
@@ -33,23 +33,25 @@ export default class Dropdown extends Component {
...
-    // then we are past the right side and need to right justify
-    const touch_right = window.innerWidth < elementRect.right;
-    // we need to push up
-    const touch_bottom = elementRect.bottom > window.innerHeight;
-
-    if (touch_bottom) {
-      // 4 because of margins
-      element.style.top = `-${(elementRect.height + 4)}px`;
-    }
+    if (window.screen.width > 479) {
```
wyseguyonline: Does this line up with a specific breakpoint that we have in Elements?

mhomolak: XS is 0-479 so yes

umahaea:
reg passed: https://travis-ci.org/Pearson-Higher-Ed/ux-test-platform/builds/303674495
good from QA

PR #99, opened 2017-11-17
fix: fixed z-index for the UL rather than the entire container

loahou04:


StommePoes:
This is probably good... just remember it if something happens where you find you can't get the dropdown part above some other area, you can probably still specifically override to set back on the container with local CSS if needed.

PR #102, opened 2017-12-08
Update radios to accept objects

jackmallahan:
There was only one test in this deprecated repository, so you might need to update tests in the SDK if you've written more.


StommePoes:
@mhomolak @wyseguyonline @umahaea 

I'm not sure the implications of swapping opt for key, other than key is inherent in react?

loahou04:
@StommePoes we need a unique identifier for the key.  However, we made it backwards compatible so that if  a user never passes in the options array as an array of objects it'll still work.  This PR didn't really change anything aside from allow it to be more generic and let the consumer set the value/label rather than just passing in a string of choices

StommePoes:
Thanks @loahou04 with that in mind, this PR sounds ok to me.

umahaea:
Looks good. good to merge

PR #103, opened 2017-12-08
fix: implement UXF dropdown in PhoneNumber

wyseguyonline:


demo/demoPages/DropdownPage.js
```diff
@@ -128,10 +135,57 @@ class DropdownPage extends React.Component {
...
+                if (data.value === 'US') {
+                  this.setState({buttonImageSelected:true});
+                }
+              }}
+              type="image"
+              label="image"
+              id="image"
+              btnImage="https://lipis.github.io/flag-icon-css/flags/4x3/us.svg"
+              btnImageHeight="10"
+              btnImageWidth="20">
```
StommePoes: I don't see where we're adding the alt attribute... it can be empty if the image is just sitting next to text anyway, but it needs to be there (and I saw it earlier in the phone number component)

demo/demoPages/PhoneNumberPage.js
```diff
@@ -1,72 +1,83 @@
...
+              <ul>
+                <li className="li-props">labelText:String === {`<PhoneNumber labelText="Your label" />`}</li>
+                <li className="li-props">Text value displayed above the Input</li>
+                <li className="li-props">infoMessage:String === {`<PhoneNumber infoMessage="How informative!" />`}</li>
+                <li className="li-props">Message displayed below the Input</li>
+                <li className="li-props">errorMessage:String === {`<PhoneNumber errorMessage="Error message" />`}</li>
+                <li className="li-props">Message displayed below the Input during an error state.  Note:  will trigger form field to be in an error state.</li>
+                <li className="li-props">country:String === {`<PhoneNumber country="US" />`}</li>
+                <li className="li-props">Two-letter country code to be used as the default country. <br/>Full list of codes can be found here&nbsp;
+                <a href="https://github.com/catamphetamine/react-phone-number-input/blob/master/source/country%20names.json" target="_blank">Country codes</a></li>
```
StommePoes: I recall locally we renamed this file to not have spaces, do we need to mention that anywhere or does it not matter to any consumers?

wyseguyonline: It shouldn't matter to consumers.  We've modified our source files to point to the file names without spaces.  But we can still point to this one (the original).

demo/demoPages/DropdownPage.js
wyseguyonline: I believe with these images, the alt text is left blank because they provide no information that a non-sighted user would need and the labeling would come from other portions of the UI.

I think we can roll in the id part into the next PR for this repo.

StommePoes: but it adds in the alt attribute? so long as it does that, we're good, because yeah, in this case the images don't do anything. I remember an alt="image" on the dropdown button tho... we don't want that.

wyseguyonline: I added alt="" to the images at the component level.  I also removed the alt="image" and replaced that as well with alt="".

StommePoes: \o/

umahaea:
Looks good merging it.

PR #104, opened 2017-12-15
Remove state from RCG, remove backwards compatibility

jackmallahan:
This removes backwards compatibility from radio check group. Now this component should only take in an array of objects. This put the responsibility on the developer to pass in objects with properties of "disabled" and/or "checked" instead of passing in a string.

 @loahou04


umahaea:
@jackmallahan  - There are some unit test failures. Pls fix it.

StommePoes:
Inputs isn't appearing on the demo page.

PR #107, opened 2017-12-21
Added props and props.children defaults to empty object getSelectedIndex

dgreenec:
Issue with undefined or empty children when setting selected index

### DIFF
--- a/src/Dropdown/Dropdown.js
+++ b/src/Dropdown/Dropdown.js
@@ -248,10 +248,11 @@ export default class Dropdown extends Component {
 
     if (this.props.children) {
       for (let i = 0; i < this.props.children.length; i++) {
-        if (this.props.children[i].props.selected) {
+        const { props = {} } = this.props.children[i] || {};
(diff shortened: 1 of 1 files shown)


src/Dropdown/Dropdown.js
```diff
@@ -248,10 +248,11 @@ export default class Dropdown extends Component {
 
     if (this.props.children) {
       for (let i = 0; i < this.props.children.length; i++) {
-        if (this.props.children[i].props.selected) {
+        const { props = {} } = this.props.children[i] || {};
```
mhomolak: Could this be changed to remove that first assignment to an empty object?

So you'd have `const { props } = this.props.children[i] || {};`

dgreenec: We have a case where without that default, props is undefined and throws an error on the next line: "TypeError: Cannot read property 'selected' of undefined".  Having the default to an empty object handles the case of this.props.children[i] being undefined.

PR #109, opened 2018-01-22
fix: Add type to Dropdown trigger

mhomolak:
- TIL: button's type defaults to submit.  Due to this it was causing PI's forms to submit.
- Also changed the parameter on the ul ref to a more appropriate name.


StommePoes:
lol
always always always always type=button, one never knows when some crazy team will put your component inside a form

PR #110, opened 2018-01-27
des-900 fix from sdk

StommePoes:
Only DES-900 was needed here, there's no progressbar nor button styles in compounds.

@mhomolak @briandunnpearson for review.

@umahaea @udhadpa this requires some changes to the tests, same as were done for static alert in elements-sdk.


umahaea:
Regression looks good. merging it. https://travis-ci.org/Pearson-Higher-Ed/ux-test-platform/builds/335234040

PR #113, opened 2018-02-15
fix: deal with weird react children

nlkluth:
Starts selectedIndex check at -1 so only valid children will get the selectedIndex value; undefined and other values are sometimes passed in to children as we build up the structure


nlkluth:
@JasonLantz @wyseguyonline @mhomolak any clue wassup with the build?

JasonLantz:
looks like it's just waiting for workers....

